PR #1, opened 2019-12-18
add handlers for language and profile and activity tracler keys info

nbirillo:
and fix project structure


src/data_handler.py
```diff
@@ -0,0 +1,45 @@
...
+
+def __get_extension(file_name):
+    parts = file_name.split(".")
+    return parts[len(parts) - 1]
+
+
+def __get_language_name(extension):
+    if consts.LANGUAGES_DICT.get(extension):
+        return consts.LANGUAGES_DICT[extension]
+    return consts.NOT_DEFINED_LANGUAGE
```
elena-lyulina: Let's use the built-in method `dict.get(key[, default])`

src/data_handler.py
```diff
@@ -0,0 +1,45 @@
...
+        return -1
+    values = np.delete(values, index)
+    if len(values) == 1:
+        return values[0]
+    return default_value
+
+
+def __get_extension(file_name):
+    parts = file_name.split(".")
+    return parts[len(parts) - 1]
```
elena-lyulina: Let's just use `parts[-1]`, it should return the last element in the list

src/data_handler.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def __get_language_name(extension):
+    if consts.LANGUAGES_DICT.get(extension):
+        return consts.LANGUAGES_DICT[extension]
+    return consts.NOT_DEFINED_LANGUAGE
+
+
+# if we have note 1 language, we return NOT_DEFINED
```
elena-lyulina: Does this comment match the function behavior?
It seems that if we have noted only one extension, we will return exactly this language:

> if len(extensions) == 1:
        return __get_language_name(extensions[0])

src/data_handler.py
```diff
@@ -0,0 +1,45 @@
...
+# Get 1 language for all files
```
elena-lyulina: Please, add more details to this description:
- which one language and in which cases
- what does 'all files' mean

src/main.py
```diff
@@ -0,0 +1,20 @@
...
+
+pd.set_option('display.max_rows', 250)
+pd.set_option('display.max_columns', 100)
+
+
+DATA_FILE_PATH = None
+AT_FILE_PATH = None
+
+
+def main():
```
elena-lyulina: Please, pass `DATA_FILE_PATH` and `AT_FILE_PATH` as arguments

src/main.py
```diff
@@ -0,0 +1,20 @@
...
+    df = pd.read_csv(DATA_FILE_PATH, encoding=consts.ENCODING, header=None)
+
+
+if __name__ == "__main__":
+    main()
+
+
+# id activity tracker, file name, time + keys (pair)
```
elena-lyulina: Maybe it's not the best place for the format of the activity-tracker data?

src/data_handler.py
nbirillo: sorry, I mean if we have NOT 1 language

nbirillo: I corrected it:  If we have a few languages

src/main.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def __index_containing_substring(list_of_string: list, substring: str):
+    for i, s in enumerate(list_of_string):
+        if substring in s:
+            return i
+    return -1
+
+
+def __get_real_ati_file_index(path: str, files: list, ati_key: str):
```
elena-lyulina: This function returns the first AT file but it doesn't check the rest of the files. It may be that something went wrong and there are several AT files. Maybe it would be better to check the rest of the files and in case of finding another one AT file throw an exception? We also don't need the __count_containing_substring and __index_containing_substring functions in this case cause we will find the real AT by __get_real_ati_file_index or throw an exception.

src/main.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def __separate_ati_and_other_files(files: list, folder: str, full_path: str):
+    count_ati = __count_containing_substring(files, consts.ACTIVITY_TRACKER_FILE_NAME)
+    ati_file = None
+    ati_id = None
+    if count_ati > 1:
+        ati_file_index = __get_real_ati_file_index(full_path, files, consts.ACTIVITY_TRACKER_FILE_NAME)
+    else:
+        ati_file_index = __index_containing_substring(files, consts.ACTIVITY_TRACKER_FILE_NAME)
```
elena-lyulina: There are other arguments of changing the __get_real_ati_file_index function and removing __count_containing_substring and __index_containing_substring:
1. in this implementation in case of only one file with AT name, this file isn't checked for having headers. For example,
there may be a CT file with events of editing an AT file, so the name of the CT file may contain AT name. 
2.  in this implementation you are iterating twice over all files (first by counting files containing AT name, second by calling  __get_real_ati_file_index or __index_containing_substring). You can avoid it by calling the changed __get_real_ati_file_index function once.

src/activity_tracker_handler.py
```diff
@@ -0,0 +1,162 @@
...
+    ati_data = ati_data[(ati_data[consts.ACTIVITY_TRACKER_COLUMN.EVENT_TYPE.value].isin(event_types))
+                        & (ati_data[consts.ACTIVITY_TRACKER_COLUMN.EVENT_DATA.value].isin(
+        consts.ACTIVITY_TRACKER_EVENTS.action_events()))]
+    return ati_data
+
+
+# Delete : symbol from hours in timestamp for correcting convert to datetime
+# For example 2019-12-09T18:41:28.548+03:00 -> 2019-12-09T18:41:28.548+0300
+def __corrected_time(timestamp: str):
+    return re.sub(r'([-+]\d{2}):(\d{2})(?:(\d{2}))?$', r'\1\2\3', timestamp)
```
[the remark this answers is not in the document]
nbirillo: I want to use the unittest module for testing, but I can not find the best way of testing the private methods (which were marked __). Because when I try to make a class for tests, I can not use the private functions from another module, I can use a global variable in current test module for calling the private functions from another module, but I think it looks not beautiful. Maybe do you know the best way of testing the private functions?
You can see the answer about it, but it doesn't help: https://softwareengineering.stackexchange.com/questions/380287/why-is-unit-testing-private-methods-considered-as-bad-practice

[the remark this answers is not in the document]
nbirillo: I made my variant tests for it. I hope it looks ok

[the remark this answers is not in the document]
elena-lyulina: Yeah, looks good, thanks!

src/main/tasks_tests_handler.py
```diff
@@ -0,0 +1,212 @@
...
+
+# For python scripts it is an in file with extension py and for other cases, it is an in file with extension txt
+def __get_in_file_for_current_test(cur_in_file: str, task: str, extension='py'):
+    if __is_python(extension):
+        __create_py_input_file(cur_in_file, task)
+        return INPUT_FILE_NAME + '.' + extension
+    return cur_in_file
+
+
+# Clear not .txt files: we need to clear the generated files (it has not .txt extension)
```
elena-lyulina: it has -> they don't have or they have no (plural, because it's about files)

elena-lyulina: I'm not sure about using the world 'clear' like this because 'clear smth' means 'keep this smth, but remove the other stuff'. So talking about old files, we don't want to clear them, we want to clear the other files from them. 
Let's change it to make it clear :)

src/main/tasks_tests_handler.py
```diff
@@ -0,0 +1,212 @@
...
+def __clear_old_files(task: str):
```
elena-lyulina: Same thing with 'clear' in the name

src/main/tasks_tests_handler.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+def __compile_program(call_args: list):
+    if call(call_args) != 0:
+        # Error
+        return False
+    return True
+
+
+def __get_compiled_file_path(extension: str, task: str, source_file_name: str):
```
elena-lyulina: The result of this function with several ifs is only used in the `__get_args_for_compiling_program` with the same ifs. Shouldn't we merge it to the `__get_args_for_compiling_program`, counting compiled file path from the right if? 
This would help us to get rid of this function that doesn't look too pretty because of so many ifs.
So, in the case of java it would be something like this:
```
if extension == 'java':
        compiled_file_path = base_path + source_file_name + '.' + extension
        call_args = ['javac', compiled_file_path]
```

src/main/tasks_tests_handler.py
```diff
@@ -0,0 +1,212 @@
...
+    files = next(os.walk(TASKS_TESTS_PATH + task))[2]
+    in_and_out_files = __get_in_and_out_files(files)
+
+    count_tests, passed_tests = len(in_and_out_files), 0
+    source_file_name = __create_source_code_file(source_code, task, extension)
+
+    if not __is_python(extension):
+        has_compiled_file, is_compiled_successful = __get_default_compiled_program_info(source_file_name, task)
+
+        compiling_args = __get_args_for_compiling_program(extension, task, source_file_name)
```
elena-lyulina: Move this string inside the following if

src/tests/activity_tracker_handler/dataframe_methods.py
```diff
@@ -0,0 +1,166 @@
...
+def get_data_for_merging_test_1():
+    ati_folder = 'ati_1'
+    ati_df = pd.read_csv(consts.TEST_DATA_PATH + '/' + ati_folder + '/' + 'ide-events_1.csv', encoding=consts.ENCODING,
+                         names=consts.ACTIVITY_TRACKER_COLUMN.activity_tracker_columns())
+    ati_df = ath.preprocessing_activity_tracker_data(ati_df)
+    ct_df = pd.read_csv(consts.TEST_DATA_PATH + '/' + ati_folder + '/' + 'task_1.csv', encoding=consts.ENCODING)
+    ct_df_right = __replace_nan_in_ati_columns(
+        pd.read_csv(consts.TEST_DATA_PATH + '/' + ati_folder + '/' + 'union_task_1.csv', encoding=consts.ENCODING))
+
+    ct_df = ath.merge_code_tracker_and_activity_tracker_data(ct_df, ati_df)
```
elena-lyulina: We should also pass an AT id here, so please, add it here, there and everywhere :)

src/main/main.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def main():
+    logging.basicConfig(filename=consts.LOGGER_FILE, level=logging.INFO)
+    log = logging.getLogger(consts.LOGGER_NAME)
+    try:
+        path = __get_data_path()
+    except NameError as e:
+        log.error(e)
+        sys.exit(1)
```
elena-lyulina: Let's also print the error message

src/main/main.py
```diff
@@ -0,0 +1,111 @@
...
+        ati_id = folder.split('_')[1]
+    return files, ati_file, ati_id
+
+
+def __get_folders_for_handling(path: str):
+    # Get child folders for the root folder from generator and remove result folder
+    return list(filter(lambda x: consts.RESULT_FOLDER not in x, next(os.walk(path))[1]))
+
+
+def __write_result(path: str, file: str, result_df: pd.DataFrame):
```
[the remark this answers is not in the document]
nbirillo: I think no because we don't have the activity tracker files when we have finished handling it and it is strange to save a folder for it or not? But of course, I can save the structure if it is necessary. Is it important to put the code tracker files from one user in the one folder?

PR #2, opened 2020-02-05
Move compiled files to a source folder

nbirillo:


elena-lyulina:
Thank you, Nastya, very cool

PR #6, opened 2020-02-12
Fix bugs for handling ati and ct files 

nbirillo:
And catch a case when a folder has a few same files


src/main/preprocessing/preprocessing.py
```diff
@@ -84,7 +99,11 @@ def preprocess_data(path):
     for folder in folders:
```
elena-lyulina: It would be great to check if we have several files with the same 'unhashed' name. In that case, we should consider at_df as None

src/main/preprocessing/activity_tracker_handler.py
```diff
@@ -173,19 +173,41 @@ def merge_code_tracker_and_activity_tracker_data(code_tracker_data: pd.DataFrame
...
+
+
+def __filter_ati_file_paths(separated_paths: list):
+    paths_dict = {}
+    for separated_path in separated_paths:
+        if paths_dict.get(__get_file_name(separated_path)) is None:
+            paths_dict[__get_file_name(separated_path)] = __get_file_path(separated_path)
+        else:
+            if paths_dict[__get_file_name(separated_path)] != __get_file_path(separated_path):
+                log.error('The folder has a few the same files!')
```
elena-lyulina: Let's say 'Activity tracker data contains several files with the same names'

src/main/preprocessing/preprocessing.py
elena-lyulina: And also let's return the new data path to the folder, all preprocessed files are written to

PR #7, opened 2020-02-17
Running task tests for task splitting

elena-lyulina:


src/main/main.py
```diff
@@ -1,129 +1,47 @@
...
-    except ValueError:
-        raise NameError('List of arguments does not have the argument -path')
-    except IndexError:
-        raise NameError('List of arguments does not have a value after the argument -path')
+    path = args[args.index(PATH_CMD_ARG) + 1]
+    if not os.path.isdir(path):
+        log.error("It's not a folder, passed path is " + path)
+        sys.exit(1)
+    path = add_slash(path)
     return path
```
nbirillo: Maybe is it better to return add_slash(path) without an additional variable path?

src/main/util/consts.py
```diff
@@ -0,0 +1,168 @@
...
+    # cin>>a;
+    # cout<<a;}
+    LANGUAGE.CPP.value: 60
+}
+
+LANGUAGE_TO_OUTPUT = {
+    LANGUAGE.PYTHON.value: "print",
+    LANGUAGE.JAVA.value: "System.out.print",
+    LANGUAGE.KOTLIN.value: "print",
+    LANGUAGE.CPP.value: "cout",
```
nbirillo: Sometimes people can use the printf function for output in the console for C++ language. You can see the documentation: http://www.cplusplus.com/reference/cstdio/printf/

src/main/util/consts.py
```diff
@@ -0,0 +1,168 @@
...
+
+DATE_TIME_FORMAT = '%Y-%m-%dT%H:%M:%S.%f%z'
+MAX_DIF_SEC = 0.5
+
+TEST_DATA_PATH = ROOT_DIR + '/../../resources/test_data'
+
+PREPROCESSING_RESULT_FOLDER = 'preprocessing_result'
+
+# v 2.0 - with stopping after the first break
+RUNNING_TESTS_RESULT_FOLDER = 'running_tests_result2'
```
nbirillo: Without 2 or with _: running_tests_result_2. But I think the running_tests_result is better

src/main/util/file_util.py
```diff
@@ -0,0 +1,147 @@
...
+def get_file_name_from_path(file_path: str, with_extension=True):
+    file_path = remove_slash(file_path)
+    file_name = file_path.split('/')[-1]
+    if not with_extension:
+        file_name = file_name.split('.')[0]
+    return file_name
+
+
+def get_extension_from_file(file: str):
+    return file.split(".")[-1]
```
nbirillo: Please use only ' symbols for strings (or " of course). But is better to use the ' symbol. It is necessary for a uniform code style. You can see for example https://stackoverflow.com/questions/628657/difference-between-the-use-of-double-quote-and-quotes-in-python

src/main/util/file_util.py
```diff
@@ -0,0 +1,147 @@
...
+        parent_folder = add_slash(parent_folder)
+    return parent_folder
+
+
+def get_parent_folder_name(file_path: str):
+    file_path = remove_slash(file_path)
+    return file_path.split('/')[-2]
+
+
+def change_extension_to(file: str, new_extension: str):
```
nbirillo: It does not work for a case when a file does not have an extension

src/main/util/file_util.py
```diff
@@ -0,0 +1,147 @@
...
+        
+def remove_directory(directory: str):
+    if os.path.exists(directory):
+        shutil.rmtree(directory, ignore_errors=True)
+
+
+# To get all files or subdirs (depends on the last parameter) from root that match item_condition
+# Can be used to get all codetracker files, all data folders, etc.
+# Note that all subdirs or files already contain the full path for them
+def get_all_file_system_items(root: str, item_condition, item_type=FILE_SYSTEM_ITEM.FILE.value):
```
nbirillo: Please use Callable type for the type annotation for item_condition. See more: https://docs.python.org/3/library/typing.html

src/main/util/file_util.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+# to get all codetracker files
+def ct_file_condition(name: str):
+    return ACTIVITY_TRACKER_FILE_NAME not in name and csv_file_condition(name)
+
+
+# to get all subdirs that contain ct and ati data
+def data_subdirs_condition(name: str):
+    return DATA_FOLDER_WITH_ATI in name or DATA_FOLDER_WITHOUT_ATI in name
```
nbirillo: Maybe rename it? Because you can mistakenly think that we take the whole space, because we use the or operator for DATA_FOLDER_WITH_ATI and DATA_FOLDER_WITHOUT_ATI

src/main/util/file_util.py
```diff
@@ -0,0 +1,147 @@
...
+
+# to get path to the result folder that is near to the original folder
+# and has the same name but with a suffix added at the end
+def get_result_folder(folder: str, result_name_suffix: str):
+    result_folder_name = get_file_name_from_path(folder) + "_" + result_name_suffix
+    return os.path.join(get_parent_folder(folder), result_folder_name)
+
+
+def check_folder_and_write_df_to_file(folder_to_write: str, file_to_write: str, df: pd.DataFrame):
+    if not os.path.exists(folder_to_write):
```
nbirillo: Please, use the create_directory function

src/main/util/file_util.py
```diff
@@ -0,0 +1,147 @@
...
+        os.makedirs(folder_to_write)
+
+    # get error with this encoding=ENCODING on ati_225/153e12:
+    # "UnicodeEncodeError: 'latin-1' codec can't encode character '\u0435' in position 36: ordinal not in range(256)"
+    # So change it then to 'utf-8'
+    try:
+        df.to_csv(file_to_write, encoding=ENCODING, index=False)
+    except UnicodeEncodeError:
+        df.to_csv(file_to_write, encoding='utf8', index=False)
```
nbirillo: Maybe should we move utf8 to the consts file?

src/main/util/file_util.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+# to write a dataframe to the result_folder remaining the same file structure as it was before
+# for example, for file path/folder1/folder2/ati_566/file.csv the dataframe will be 
+# written to result_folder/folder1/folder2/ati_566/file.csv
+def write_result(result_folder: str, path: str, file: str, df: pd.DataFrame):
+    path_from_result_folder_to_file = file[len(path):]
```
nbirillo: You should check not only the length of the path but also a coincidence of the common part because we can have the same length for two paths, but it may be the different parts

src/main/util/consts.py
elena-lyulina: Yeah, it's better, but I used the 'running_test_result' in my previous version, the one without stopping after the first false getting in tests. Since the results of these two versions are different, we don't want to write them in the same folder. Although, I can add '_' if you want

src/main/splitting/tasks_tests_handler.py
```diff
@@ -0,0 +1,290 @@
...
+
+def __run_python_test(in_file: str, expected_out: str, task: str, source_file_name=SOURCE_OBJECT_NAME):
+    p1 = Popen(['python3', __get_task_file(in_file, task)], stdout=PIPE)
+    p2 = Popen(['python3', __get_compiled_file(source_file_name) + '.'
+                + get_extension_by_language(LANGUAGE.PYTHON.value)],
+               stdin=p1.stdout,
+               stdout=PIPE)
+    p1.stdout.close()
+    try:
+        signal.alarm(consts.MAX_SECONDS_TO_WAIT_TEST)
```
nbirillo: I think we need to stop alarm (see signal.alarm(0)). Because we can get incorrect results, see more: https://stackoverflow.com/questions/27013127/stop-sigalrm-when-function-returns

src/main/splitting/tasks_tests_handler.py
```diff
@@ -0,0 +1,290 @@
...
+from src.main.util.file_util import get_content_from_file, create_file, create_directory, remove_directory
+
+INPUT_FILE_NAME = consts.TASKS_TESTS.INPUT_FILE_NAME.value
+TASKS_TESTS_PATH = consts.TASKS_TESTS.TASKS_TESTS_PATH.value
+SOURCE_OBJECT_NAME = consts.TASKS_TESTS.SOURCE_OBJECT_NAME.value
+
+log = logging.getLogger(consts.LOGGER_NAME)
+
+
+class TimeoutException(Exception):
```
nbirillo: Maybe should we move the TimeoutException class and the timeout_handler function into a separate file? Because we can reuse it in the future in other places

src/main/splitting/tasks_tests_handler.py
```diff
@@ -0,0 +1,290 @@
...
+    try:
+        signal.alarm(consts.MAX_SECONDS_TO_WAIT_TEST)
+        out, err = p2.communicate()
+        actual_out = out.decode("utf-8").rstrip("\n")
+        log.info("In-file: " + in_file + ", task: " + task + ", expected out: " + expected_out + ", actual out: " + actual_out)
+        return actual_out == expected_out
+    except TimeoutException:
+        log.info("In-file: " + in_file + ", task: " + task + ", Time is out")
+        return False
+    except:
```
nbirillo: Please fix it. It is a bad way - to use only 'except' without a type of error. You can read about it here: https://stackoverflow.com/questions/21553327/why-is-except-pass-a-bad-programming-practice/21553515. If you want to catch any "safe" exceptions, please use 'except Exception', see more: https://stackoverflow.com/questions/730764/how-to-properly-ignore-exceptions

src/main/splitting/tasks_tests_handler.py
```diff
@@ -0,0 +1,290 @@
...
+    if language == LANGUAGE.JAVA.value:
+        source_file_name = __get_java_class(source_code)
+    create_file(source_code, get_extension_by_language(language), __get_compiled_file(source_file_name))
+    return source_file_name
+
+
+def create_in_and_out_dict(tasks: list):
+    in_and_out_files_dict = {}
+    for task in tasks:
+        files = next(os.walk(TASKS_TESTS_PATH + task))[2]
```
nbirillo: Replace please 2 to FILE_SYSTEM_ITEM.FILE.value from the consts file

src/main/splitting/tasks_tests_handler.py
```diff
@@ -0,0 +1,290 @@
...
+            return True
+        except:
+            return False
+    return False
+
+
+def check_python_file_by_mypy(file_name: str):
+    call_args = ['mypy', file_name]
+    is_valid = __compile_program(call_args)
+    return is_valid
```
nbirillo: Delete the additional variable is_valid. We should use: return __compile_program(call_args)

src/main/splitting/tasks_tests_handler.py
```diff
@@ -0,0 +1,290 @@
...
+    # not to check incorrect fragments
+    if not is_source_file_correct(source_file, language):
+        need_to_run_tests = False
+        rate = consts.TEST_RESULT.INCORRECT_CODE.value
+        test_results = [rate] * len(tasks)
+
+    # not to check too small fragments because they cannot return true anyway
+    elif len(source_code) < consts.LANGUAGE_TO_MIN_SYMBOLS[language]:
+        log.info("Code fragment is too small")
+        need_to_run_tests = False
```
nbirillo: I think we can define a function for the code fragment:
`need_to_run_tests = False`
`test_results = [rate] * len(tasks)`
Because we use it 3 times and it is "default values" for these cases

src/main/splitting/splitting.py
```diff
@@ -0,0 +1,106 @@
...
+    language = get_ct_language(data)
+    # add unique to logs
+    log.info(file_log_info + ", language is " + language + ", found " + str(data.shape[0]) + " fragments")
+
+    if language is not consts.LANGUAGE.NOT_DEFINED.value:
+        unique_fragments = list(data[FRAGMENT].unique())
+        log.info("Found " + str(len(unique_fragments)) + " unique fragments")
+
+        fragment_to_test_results_dict = dict(map(lambda f: (f, check_tasks(tasks, f, in_and_out_files_dict, language)), unique_fragments))
+        data[consts.CODE_TRACKER_COLUMN.TESTS_RESULTS.value] = data.apply(lambda row:
```
nbirillo: Why do you use the variable FRAGMENT, which you defined in the top of the file, but consts.CODE_TRACKER_COLUMN.TESTS_RESULTS.value? I think you should define column's names in the top of the file for all cases or in all cases use the full version

src/main/splitting/splitting.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def check_tasks_on_correct_fragments(data: pd.DataFrame, tasks: list, in_and_out_files_dict: dict, file_log_info=""):
+    data[FRAGMENT] = data[FRAGMENT].fillna("")
+    # if run after preprocessing, this value can be taken from 'language' column
+    language = get_ct_language(data)
+    # add unique to logs
+    log.info(file_log_info + ", language is " + language + ", found " + str(data.shape[0]) + " fragments")
+
+    if language is not consts.LANGUAGE.NOT_DEFINED.value:
```
nbirillo: It is difficult case when we have two NOT - is not and NOT DEFINED. Could you please swap if and else bodies?
`if language is consts.LANGUAGE.NOT_DEFINED.value: actions`
`else: actions`

src/main/splitting/splitting.py
```diff
@@ -0,0 +1,106 @@
...
+        return real_splits
+
+    prev_split = supposed_splits[0]
+    prev_intersected_tasks = prev_split[consts.SPLIT_DICT.TASKS.value]
+
+    for curr_split in supposed_splits:
+        curr_intersected_tasks = intersect(prev_intersected_tasks, curr_split[consts.SPLIT_DICT.TASKS.value])
+        if len(curr_intersected_tasks) == 0:
+            # it means that the supposed task has changed, so we should split on prev_split
+            real_splits.append({consts.SPLIT_DICT.INDEX.value: prev_split[consts.SPLIT_DICT.INDEX.value],
```
nbirillo: Can we move it in the separate function to add some values in the dictionary? Dictionary passed by links in python

src/main/splitting/splitting.py
```diff
@@ -0,0 +1,106 @@
...
+                        consts.SPLIT_DICT.RATE.value: prev_split[consts.SPLIT_DICT.RATE.value],
+                        consts.SPLIT_DICT.TASKS.value: prev_intersected_tasks})
+    return real_splits
+
+
+def filter_already_tested_files(files: list, result_folder_path: str):
+    tested_files = get_all_file_system_items(result_folder_path, ct_file_condition, consts.FILE_SYSTEM_ITEM.FILE.value)
+    # to get something like 'ati_239/Main_2323434_343434.csv'
+    tested_folder_and_file_names = list(map(lambda f: get_parent_folder_name(f) + "/" + get_file_name_from_path(f), tested_files))
+    return list(filter(lambda f: get_parent_folder_name(f) + "/" + get_file_name_from_path(f) not in tested_folder_and_file_names, files))
```
nbirillo: I think we can define an internal function, which calculates `get_parent_folder_name(f) + "/" + get_file_name_from_path(f)` and use it in the 79 and 80 rows

src/main/util/consts.py
nbirillo: I think it is better because we use _ for separating parts in all names in our project

src/main/splitting/task_checker.py
```diff
@@ -0,0 +1,144 @@
...
+        elif not does_string_contain_any_of_substrings(source_code, self.output_strings):
+            log.info('Code fragment does not contain any output strings')
+            need_to_run_tests, test_results = self.get_no_need_to_run_tests_values(rate, len(tasks))
+
+        return need_to_run_tests, test_results, rate
+
+    def check_task(self, task: str, in_and_out_files_dict: dict, source_file: str, stop_after_first_false=True):
+        log.info(f'Start checking task {task}')
+        in_and_out_files = in_and_out_files_dict.get(task)
+        if in_and_out_files is None:
```
nbirillo: I think we should check if len(in_and_out_files) is zero, and if it is, we should raise an error because it is our mistake (we forgot about in and out files for the task and then we will have the division by zero error during rate calculating further). Maybe it is better the check of it to put in the task_tests_handler file because it is an init state

src/main/splitting/task_checker.py
```diff
@@ -0,0 +1,144 @@
...
+
+        counted_tests, passed_tests = len(in_and_out_files), 0
+        for in_file, out_file in in_and_out_files:
+            is_passed = self.run_test(get_content_from_file(in_file), get_content_from_file(out_file), source_file)
+            log.info(f'Test {in_file} for task {task} is passed: {str(is_passed)}')
+            if is_passed:
+                passed_tests += 1
+            elif stop_after_first_false:
+                log.info('Stop after first false')
+                break
```
nbirillo: Should we make the variable passed_tests = 0 before the break? Otherwise, we will have an incorrect result, because it can be passed some tests, passed_tests will be not zero and our rate will be not zero, but maybe other tests would be passed and the right value of rate must have to other value. Or I don't fully understand fully this logic

src/main/splitting/python_task_checker.py
```diff
@@ -0,0 +1,44 @@
...
+    @property
+    def min_symbols_number(self):
+        return 20
+
+    @property
+    def output_strings(self):
+        return ['print']
+
+    def create_source_file(self, source_code: str):
+        source_code_file = os.path.join(TASKS_TESTS_PATH, SOURCE_OBJECT_NAME,
```
nbirillo: We have the same the code fragment in all checkers (we create source_code_file path and after to run create_file). I think it would be better to put it in a separate static method and run in each class with necessary parameters

src/main/splitting/python_task_checker.py
```diff
@@ -0,0 +1,37 @@
...
+
+    @property
+    def output_strings(self):
+        return ['print']
+
+    def create_source_file(self, source_code: str):
+        return self.create_source_file_with_name(source_code, SOURCE_OBJECT_NAME)
+
+    def is_source_file_correct(self, source_file: str):
+        is_correct = check_call_safely(['mypy', source_file]) and check_call_safely([sys.executable, source_file])
```
nbirillo: Should we check if mypy installed?

PR #8, opened 2020-02-28
Add statistic gathering and plots with plotty for statistic

nbirillo:


src/main/plots/save_statistics.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def __get_statistics_info_for_plots(column: stat_const.STATISTICS_COLUMN):
+    return {
+        stat_const.STATISTICS_INFO_FOR_PLOTS.LABELS.value: __get_labels_for_plots(column),
+        stat_const.STATISTICS_INFO_FOR_PLOTS.TITLE.value: __get_title_for_plots(column)
+    }
+
+
+def plot_statistics(file: str, column: stat_const.STATISTICS_COLUMN, plot_type: stat_const.PLOT_TYPES,
```
elena-lyulina: It fails when we pass STATICTIC_COLUMN.FREQ.value in 'column' argument. Maybe it would be better to create a separate variable for FREQ and remove it from STATICTIC_COLUMN?

src/main/plots/consts.py
```diff
@@ -23,3 +26,47 @@
...
+    BAR = 'bar'
+
+
+class STATISTICS_INFO_FOR_PLOTS(Enum):
+    LABELS = 'labels'
+    TITLE = 'title'
+
+
+class OUTPUT_FORMAT(Enum):
+    HTML = 'html'
```
elena-lyulina: Even if it works without the dot, let's add a dot to the extension to match file_util format?

src/main/plots/save_statistics.py
nbirillo: I have changed STATICTIC_COLUMN to STATISTIC_KEY and add a new value STATISTIC_FREQ for FREQ

src/main/statistics_gathering/get_statistics.py
```diff
@@ -0,0 +1,117 @@
...
+    get_result_folder, create_file, change_extension_to
+from src.main.preprocessing.preprocessing import __separate_ati_and_other_files
+from src.main.preprocessing.code_tracker_handler import handle_ct_file
+
+
+log = logging.getLogger(consts.LOGGER_NAME)
+
+
+# We must have one value in a profile column else it is an incorrect case
+def __get_profile_info(ct_df: pd.DataFrame, column: consts.CODE_TRACKER_COLUMN):
```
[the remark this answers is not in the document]
nbirillo: No, handle_ct_file, in this case, returns -1, because it is the incorrect case for us (it is a case from STATISTICS_SHOWING_KEY - INCORRECT) and we will have  always  only 1 value in the one file

src/main/statistics_gathering/get_statistics.py
```diff
@@ -0,0 +1,117 @@
...
+    experience = __get_profile_info(ct_df, consts.CODE_TRACKER_COLUMN.EXPERIENCE.value)
+    log.info(f'File: {ct_file}, age is {age}, experience is {experience}')
+    return age, experience
+
+
+# Handle set of profile data values
+# Return default_value if files with the same code tracker id (or the same activity tracker id) have differents values
+# for profile data (age or experience for example)
+# Note: you should run it for each profile column
+def __handle_profile_data_of_one_student(profile_data: set, default_value=None):
```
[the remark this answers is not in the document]
nbirillo: I think no, I have written about it above

elena-lyulina:
Looks great, thank you!

PR #9, opened 2020-03-02
Task splitting

elena-lyulina:
Add task splitting; 
Create plots to compare original splits as users marked them and real splits based on tests;
Add writing split data frames into separate files according to chosen tasks.


src/main/splitting/splitting.py
```diff
@@ -1,117 +1,76 @@
...
-        data[TESTS_RESULTS] = str([consts.TEST_RESULT.LANGUAGE_NOT_DEFINED.value] * len(tasks))
+TESTS_RESULTS = consts.CODE_TRACKER_COLUMN.TESTS_RESULTS.value
+CHOSEN_TASK = consts.CODE_TRACKER_COLUMN.CHOSEN_TASK.value
+TASK_STATUS = consts.CODE_TRACKER_COLUMN.TASK_STATUS.value
+
+
+def get_solved_task(tests_results: str):
+    tests_results = list(map(int, ast.literal_eval(tests_results)))
+    tasks = consts.TASK.tasks()
+    solved_tasks = [t for i, t in enumerate(tasks) if tests_results[i] == 1]
```
nbirillo: I think it would be better if you compare len(tasks) and len(tests_results). Because you use an index i for tasks, but if tests_results contains fewer elements than tasks we should raise an error because it is an incorrect case

src/main/splitting/splitting.py
```diff
@@ -1,117 +1,76 @@
...
-
-    if language is consts.LANGUAGE.NOT_DEFINED.value:
-        data[TESTS_RESULTS] = str([consts.TEST_RESULT.LANGUAGE_NOT_DEFINED.value] * len(tasks))
+TESTS_RESULTS = consts.CODE_TRACKER_COLUMN.TESTS_RESULTS.value
+CHOSEN_TASK = consts.CODE_TRACKER_COLUMN.CHOSEN_TASK.value
+TASK_STATUS = consts.CODE_TRACKER_COLUMN.TASK_STATUS.value
+
+
+def get_solved_task(tests_results: str):
+    tests_results = list(map(int, ast.literal_eval(tests_results)))
```
nbirillo: I think we should use tests results = ast.literal_eval(tests results) because it returns an array, which contains int and float values. Maybe it will be important in the future

src/main/splitting/splitting.py
```diff
@@ -1,117 +1,76 @@
...
+    tasks = consts.TASK.tasks()
+    solved_tasks = [t for i, t in enumerate(tasks) if tests_results[i] == 1]
+    if len(solved_tasks) == 0:
+        log.info(f'No solved tasks found, tests results: {tests_results}')
+        return np.nan
```
nbirillo: I think it would be better to move it in const file

src/main/splitting/splitting.py
```diff
@@ -1,117 +1,76 @@
...
-        file_log_info = f'file: {str(i + 1)}/{str_len_files}'
-        log.info(f'Start running tests on {file_log_info}, {file}')
-        data = pd.read_csv(file, encoding=consts.ISO_ENCODING)
-        language, data = check_tasks_on_correct_fragments(data, tasks, in_and_out_files_dict, file_log_info)
-        log.info(f'Finish running tests on {file_log_info}, {file}')
-        write_based_on_language(result_folder, file, data, language)
-
-    return result_folder
+        log.error(f'Several tasks are solved: {solved_tasks}, tests results: {tests_results}')
+        raise ValueError(f'Several tasks are solved: {solved_tasks}, tests results: {tests_results}')
```
nbirillo: Do we really want to throw an error here? It will stop handler for all files. Maybe it is better to use some special value for this case?

src/main/splitting/splitting.py
```diff
@@ -1,117 +1,76 @@
...
+
+
+def find_splits(ct_df: pd.DataFrame):
+    # fill chosen task according to solved task
+    ct_df[CHOSEN_TASK] = ct_df.apply(lambda row: get_solved_task(row[TESTS_RESULTS]), axis=1)
+
+    # change task status according to chosen task
+    ct_df.loc[ct_df[CHOSEN_TASK].isnull(), TASK_STATUS] = np.nan
```
nbirillo: I think we should add a new value to the `сonsts.TASK_STATUS` for this case and use it here. What do you think?

src/main/splitting/splitting.py
```diff
@@ -1,117 +1,76 @@
...
+    ct_df.loc[ct_df[CHOSEN_TASK].notnull(), TASK_STATUS] = consts.TASK_STATUS.SOLVED.value
+
+    # backward fill chosen task
+    ct_df[CHOSEN_TASK] = ct_df[CHOSEN_TASK].bfill()
```
nbirillo: It is a good choice!

src/main/splitting/splitting.py
```diff
@@ -1,117 +1,76 @@
...
+        for task in TASK.tasks():
+            task_df = split_df.loc[split_df[CHOSEN_TASK] == task]
+            if not task_df.empty:
+                # change name to get something like pies/ati_207_test_5894859.csv
+                filename = task + '/' + get_parent_folder_name(file) + '_' + get_name_from_path(file)
+                write_based_on_language(result_folder, filename, task_df, language)
+
+
+# run after 'split_tasks_into_separate_files' to print simple statistic
+def get_tasks_statistic(path: str):
```
nbirillo: I think we should move it to the statistics gathering folder

src/main/plots/plots_common.py
```diff
@@ -0,0 +1,60 @@
...
+EVENT_TYPE_COL = consts.ACTIVITY_TRACKER_COLUMN.EVENT_TYPE.value
+EVENT_DATA_COL = consts.ACTIVITY_TRACKER_COLUMN.EVENT_DATA.value
+
+log = logging.getLogger(consts.LOGGER_NAME)
+
+
+def get_short_name(path: str):
+    folder = get_parent_folder_name(path)
+    file_name = get_name_from_path(path)
+    return os.path.join(folder, file_name[:plot_consts.SHORT_NAME_LENGTH] + '...' + file_name[-plot_consts.SHORT_NAME_LENGTH:])
```
nbirillo: It would be better if we move it a separate function, for example:
`def string_crop(string: str, short_part_length: int, crop_symbol='...'):
    return ''.join((string[:short_part_length], crop_symbol, string[-short_part_length:]))`
(we should still think about the names)
And maybe move it to the string util module

src/main/plots/plots_common.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def save_plot(folder_to_save: str, data_path: str, fig: plt.figure, name_prefix: str):
```
nbirillo: I think we should discuss about structure of plots folder because statistics_plots file also has __save_plot method, but it is different since the plottly library is used there

src/main/plots/plots_common.py
```diff
@@ -0,0 +1,60 @@
...
+    log.info('Saving' + data_path)
+    name = name_prefix + '_' + (get_file_and_parent_folder_names(change_extension_to(data_path, EXTENSION.PNG.value)).replace('/', '_'))
```
nbirillo: Please put the format variable in the function parameters with default value EXTENSION.PNG.value

src/main/plots/plots_common.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+# add fragments lengths to the plot
+def add_fragments_length_plot(ax: plt.axes, data: pd.DataFrame):
+    ax.scatter(data[TIMESTAMP_COL], data[plot_consts.FRAGMENT_LENGTH_COL],
+               color=plot_consts.FRAGMENT_LENGTH_COLOR, s=plot_consts.SMALL_SIZE)
+
+
+# to remove previously saved plots
+def remove_all_png_files(root: str):
```
nbirillo: Maybe should we move it to the file util module?

src/main/plots/plots_common.py
```diff
@@ -0,0 +1,60 @@
...
+    files = get_all_file_system_items(root, png_file_condition, consts.FILE_SYSTEM_ITEM.FILE.value)
+    for file in files:
+        remove_file(file)
+
+
+# these functions are useful to analyze data and find files to make plots based on them
+def print_all_unique_values_in_file(file: str, column: str):
```
nbirillo: Is it helper function? Because it does not return any information and print some info only` if ati_events`

src/main/plots/ati_data_plots.py
```diff
@@ -0,0 +1,44 @@
...
+from src.main.plots import consts as plot_consts
+from src.main.plots.plots_common import save_plot, add_fragments_length_plot, get_short_name, EVENT_DATA_COL, \
+    TIMESTAMP_COL, FRAGMENT_COL
+
+log = logging.getLogger(consts.LOGGER_NAME)
+
+
+def __create_ati_events_plot(ax: plt.axes, df: pd.DataFrame, event_data: list, event_colors: dict, title: str):
+    add_fragments_length_plot(ax, df)
+    for event_data in event_data:
```
nbirillo: Rename it please, because `for event_data in event_data` looks not good

src/main/plots/ati_data_plots.py
```diff
@@ -0,0 +1,44 @@
...
+        event_data_df = df.loc[df[EVENT_DATA_COL] == event_data]
+        ax.scatter(event_data_df[TIMESTAMP_COL], event_data_df[plot_consts.FRAGMENT_LENGTH_COL],
```
nbirillo: It is very similar to `add_fragments_length_plot` from the `plots_common` file. Maybe should we change `add_fragments_length_plot`: add default parametrs `color=plot_consts.FRAGMENT_LENGTH_COLOR`, `s=plot_consts.SMALL_SIZE` and empty label?

src/main/plots/ati_data_plots.py
```diff
@@ -0,0 +1,44 @@
...
+                   color=event_colors[event_data], s=plot_consts.LARGE_SIZE, label=event_data)
+    ax.legend(bbox_to_anchor=(1.04, 1), borderaxespad=0)
```
nbirillo: Could you please add some description on the comments for different params like bbox_to_anchor and others? Like in statistics_plots file: `# x_category_order='total ascending' means: in order of increasing values in Y`. Because we use different libraries and it’s not always easy to find why exactly such parameters were used. Also, I think it would be useful in the future to quickly remember what these options mean

src/main/plots/ati_data_plots.py
```diff
@@ -0,0 +1,44 @@
...
+    ax.set_title(title)
+
+
+# create plots with different event types (running events and editor events), taken from ati data
+def create_ati_data_plot(path: str, folder_to_save=None, to_show=False):
+    data = pd.read_csv(path, encoding=consts.ISO_ENCODING)
+    data[plot_consts.FRAGMENT_LENGTH_COL] = data[FRAGMENT_COL].fillna('').str.len()
+
+    fig, (ax_run, ax_editor) = plt.subplots(2, 1, figsize=(20, 10))
+    run_title = f'run events in {get_short_name(path)}'
```
nbirillo: Maybe is it better to capitalize the first letter in the title? And below too

src/main/plots/splitting_plots.py
```diff
@@ -0,0 +1,86 @@
...
+    data[plot_consts.FRAGMENT_LENGTH_COL] = data[FRAGMENT_COL].fillna('').str.len()
+    if to_snake_case:
+        data[CHOSEN_TASK_COL] = data[CHOSEN_TASK_COL].fillna('').apply(lambda t: convert_camel_case_to_snake_case(t))
+    if not task_split_x_dict:
+        task_split_x_dict = __create_task_split_x_dict(data)
+    if not status_x_dict:
+        status_x_dict = __create_status_x_dict(data)
+
+    # add colored background to the plot
+    max_fragment_length = data[plot_consts.FRAGMENT_LENGTH_COL].max()
```
nbirillo: Maybe move it to the separate function? Also, add task status to the plot too

src/main/plots/splitting_plots.py
```diff
@@ -0,0 +1,86 @@
...
+    real_splits_data = find_splits(data)
+    fig, (ax, real_splits_ax) = plt.subplots(2, 1, figsize=(20, 10))
+
+    title = f'{get_short_name(path)} with original splits'
+    __create_splitting_plot(ax, data, title, to_snake_case=to_snake_case)
+
+    real_splits_title = f'{get_short_name(path)} with real splits'
+    __create_splitting_plot(real_splits_ax, real_splits_data, real_splits_title, to_snake_case=to_snake_case)
+
+    if folder_to_save:
```
nbirillo: You have the same fragment (in the ati_data_plots). Maybe is it also can we move to the plots common file?

src/main/splitting/splitting.py
elena-lyulina: Yes, it will, but I thought that having two tasks solved is an incorrect case, meaning that we have a mistake in run_tests function or anywhere else. So, we cannot continue data handling anyway. Having some special value for this case instead of error throwing forces us to do an additional check for this value after splitting and stop everything in case of a successful check. 
However, this process of splitting takes not so many time to be worried about stopping, and I think that throwing an error is the best way to know about errors asap.

src/main/splitting/splitting.py
elena-lyulina: Good idea, thanks!

PR #10, opened 2020-03-05
Add tasks statistics plots

nbirillo:
Add task statistics plots
Add plotty util


src/main/statistics_gathering/statistics_gathering.py
```diff
@@ -108,13 +108,24 @@ def get_statistics(path: str):
...
 # run after 'split_tasks_into_separate_files' to print simple statistic
 def get_tasks_statistic(path: str):
+    statistic = {}
     languages = [l.value for l in consts.LANGUAGE]
     language_folders = get_all_file_system_items(path, (lambda f: does_string_contain_any_of_substrings(f, languages)),
                                                  consts.FILE_SYSTEM_ITEM.SUBDIR.value)
     for l_f in language_folders:
-        print(get_name_from_path(l_f, False))
+        language = consts.LANGUAGE(get_name_from_path(l_f, False))
+        if statistic.get(language) is not None:
```
elena-lyulina: There is another way to check if variable is not None:
`if not statistic.get(language)`
instead of
`if statistic.get(language) is not None:`
because of python typecast to boolean

For me, the first way is more readable because the second has 'double negative': not and None. However, you can leave it as you prefer, I just want you to be aware of the other option.

src/main/plots/tasks_statistics_plots.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def __get_colors():
+    colors = plot_consts.BAR_PALETTE
+    # At the end of the list we have more dark colours
+    colors.reverse()
+    return colors
+
+
+def __plot_bar_chart(bars: list, path: str, plot_name='result_plot', format=plot_consts.OUTPUT_FORMAT.HTML.value,
```
elena-lyulina: Why this plot has the name 'result_plot'? Maybe rename it to 'task_plot'? Or something relative to tasks and language distribution?

src/main/plots/plotly_util.py
```diff
@@ -0,0 +1,16 @@
...
+from src.main.plots import consts as plot_consts
+from src.main.util.file_util import change_extension_to
+from src.main.plots.plots_common import create_directory_for_plots
+
+
+log = logging.getLogger(consts.LOGGER_NAME)
+
+
+def save_plot(fig, path: str, plot_type: plot_consts.PLOT_TYPES, plot_name='result_plot',
+              format=plot_consts.OUTPUT_FORMAT.HTML.value, auto_open=False):
```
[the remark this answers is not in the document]
nbirillo: Yes, I have changed it to consts.EXTENSION.HTML, thanks!

src/main/statistics_gathering/statistics_gathering.py
```diff
@@ -108,13 +108,24 @@ def get_statistics(path: str):
 
 # run after 'split_tasks_into_separate_files' to print simple statistic
 def get_tasks_statistic(path: str):
+    statistic = {}
```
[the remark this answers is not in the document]
nbirillo: O, thanks a lot! I have not known the difference

src/main/statistics_gathering/statistics_gathering.py
nbirillo: Really better, thanks!

elena-lyulina:
Great, thanks a lot  :)

PR #14, opened 2020-03-16
Solution space graph

elena-lyulina:


src/main/solution_space/solution_graph.py
```diff
@@ -52,3 +58,97 @@ def add_child(self, child: 'Vertex') -> None:
...
+        if code_info_chain:
+            log.info(f'Connect the first vertex in a chain to the start vertex')
+            first_code_info = code_info_chain[0]
+            first_vertex = self.find_or_create_vertex(first_code_info[0], first_code_info[1])
+            self.connect_to_start_vertex(first_vertex)
+
+            prev_vertex = first_vertex
+            for next_code_user in code_info_chain[1:]:
+                next_vertex = self.find_or_create_vertex(next_code_user[0], next_code_user[1])
+                prev_vertex.add_child(next_vertex)
```
nbirillo: Should we use `next_vertex.add_parent(prev_vertex)` if we have created a new vertex? It is never used and we lose some links.
Moreover I think we should call  `add_parent` and `add_child` functions in the `find_or_create_vertex ` function, because it is the  internal actions

src/main/solution_space/solution_space_handler.py
```diff
@@ -19,6 +20,14 @@ def __get_column_value(solutions: pd.DataFrame, index: int, column=None) -> str:
     return solutions[column].iloc[index]
 
 
+def __get_column_unique_value(solutions: pd.DataFrame, column=None) -> str:
+    unique_values = solutions[column].unique()
+    if len(unique_values) != 1:
+        log.error(f'There is more than 1 unique value in column {column}: {unique_values}')
+        raise ValueError(f'There is more than 1 unique value in column {column}: {unique_values}')
+    return unique_values[0]
```
nbirillo: If the column has 0 unique values we will have an error

src/main/solution_space/solution_space_handler.py
```diff
@@ -19,6 +20,14 @@ def __get_column_value(solutions: pd.DataFrame, index: int, column=None) -> str:
     return solutions[column].iloc[index]
```
nbirillo: The function can return not only str

src/main/solution_space/solution_space_handler.py
```diff
@@ -19,6 +20,14 @@ def __get_column_value(solutions: pd.DataFrame, index: int, column=None) -> str:
...
 
 
+def __get_column_unique_value(solutions: pd.DataFrame, column=None) -> str:
```
nbirillo: The function can return not only str:  the `unique()` function saves type for values in column

src/main/solution_space/solution_space_handler.py
```diff
@@ -54,17 +63,30 @@ def __find_same_fragments(solutions: pd.DataFrame, start_index: int) -> Tuple[in
...
 
 
-def __get_profile(solutions: pd.DataFrame, index: int) -> Profile:
-    age = __get_column_value(solutions, index, consts.CODE_TRACKER_COLUMN.AGE.value)
-    experience = __get_column_value(solutions, index, consts.CODE_TRACKER_COLUMN.EXPERIENCE.value)
+def __get_experience_from_str(experience: str) -> Union[EXPERIENCE, DEFAULT_VALUES]:
+    for e in EXPERIENCE:
+        if e.value == experience:
+            return e
+    return DEFAULT_VALUES.EXPERIENCE
```
nbirillo: You forgot about `value` attribute: `return DEFAULT_VALUES.EXPERIENCE.value` and  `return e.value` too

src/main/solution_space/data_classes.py
```diff
@@ -30,34 +31,51 @@ def is_empty(self) -> bool:
 
 
 class Profile:
-    def __init__(self, age=consts.DEFAULT_VALUES.AGE.value, experience=consts.DEFAULT_VALUES.EXPERIENCE.value):
+    def __init__(self, age=consts.DEFAULT_VALUES.AGE.value, experience: Union[EXPERIENCE, DEFAULT_VALUES] = DEFAULT_VALUES.EXPERIENCE):
```
nbirillo: Use type annotation in all cases or don't use in all cases

src/main/solution_space/solution_graph.py
```diff
@@ -52,3 +58,97 @@ def add_child(self, child: 'Vertex') -> None:
     def add_parent(self, parent: 'Vertex') -> None:
         self.__add_parent_to_list(parent)
         parent.__add_child_to_list(self)
+
+    def add_code_info(self, code_info: CodeInfo) -> None:
+        self._code_infos.append(code_info)
```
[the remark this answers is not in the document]
elena-lyulina: I think, non-unique elements can be there, for example, if a user was in this state twice. In this case we should add them twice, but when we call get_unique_users we'll have the actual user's number

src/main/solution_space/solution_graph.py
```diff
@@ -52,3 +58,97 @@ def add_child(self, child: 'Vertex') -> None:
...
+    def get_traversal(self):
+        return self.__iter__().traversal
+
+    def find_or_create_vertex(self, code: Optional[Code], code_info: CodeInfo) -> Vertex:
+        if code is None:
+            log.info('Code should not be None')
+            raise ValueError('Code should not be None')
+        vertices = iter(self)
+        for vertex in vertices:
+            if vertex.code and are_asts_equal(vertex.code.ast, code.ast):
```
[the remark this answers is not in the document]
elena-lyulina: Yeah, the start vertex that we decide should be in vertices has None. However, I can remove start vertex from the list and not check for None anymore. What do you think?

src/main/solution_space/data_classes.py
elena-lyulina: I'll do it later when we find out what to do with enums :)

src/main/solution_space/solution_space_handler.py
```diff
@@ -1,28 +1,50 @@
...
+    if len(unique_values) > 1:
+        log.error(f'There is more than 1 unique value in column {column}: {unique_values}')
+        raise ValueError(f'There is more than 1 unique value in column {column}: {unique_values}')
+    return unique_values[0]
 
-def __get_column_value(solutions: pd.DataFrame, index: int, column=None) -> str:
-    return solutions[column].iloc[index]
+
+def __get_enum_or_default(enum_meta: enum.EnumMeta, value: str, default: DEFAULT_VALUES) -> Union[enum.EnumMeta, DEFAULT_VALUES]:
+    for event in enum_meta:
```
nbirillo: Please change it to: `EnumMeta._value2member_map_.get(value, None)`

PR #15, opened 2020-03-16
Add methods for file cropping

nbirillo:


src/test/util/data_util/data_util_tests.py
```diff
@@ -0,0 +1,64 @@
...
+
+def get_data_with_deleted_by_start() -> Tuple[pd.DataFrame, int, None]:
+    start = 5
+    timestamps = range(start, END_INDEX)
+    return pd.DataFrame({
+        COLUMN.value: timestamps
+    }), start, None
+
+
+def are_equal(df_1: pd.DataFrame, df_2: pd.DataFrame):
```
elena-lyulina: Please, add return type :) Maybe I should worry less about typing, but from my point of view it makes our code looks beautiful 🌺

PR #17, opened 2020-03-18
Add second version of dataframe splitting into tasks

elena-lyulina:


src/main/solution_space/gum_tree.py
```diff
@@ -0,0 +1,204 @@
+#!/usr/bin/env python3
```
nbirillo: Please, add copyright to this file:  # Copyright (c) 2020 Dmitry Abramov

src/main/util/consts.py
```diff
@@ -91,10 +91,14 @@ class TASK(Enum):
     MAX_DIGIT = 'max_digit'
 
     @classmethod
-    def tasks(cls):
+    def tasks_values(cls):
         return [cls.PIES.value, cls.MAX_3.value, cls.ZERO.value,
```
nbirillo: Maybe is it better to use `__members__` function? Is it right that we want to get all values from TASK? It is better to use `[member.value for _, member in TASK.__members__.items()]` and we may not change  the code after adding a new element in the TASK

src/main/util/consts.py
```diff
@@ -91,10 +91,14 @@ class TASK(Enum):
...
                 cls.ELECTION.value, cls.BRACKETS.value, cls.MAX_DIGIT.value]
 
+    @classmethod
+    def tasks(cls):
+        return [cls.PIES, cls.MAX_3, cls.ZERO, cls.ELECTION, cls.BRACKETS, cls.MAX_DIGIT]
```
nbirillo: I think it is better to use the loop: `[task for task in TASK]`

src/main/splitting/splitting.py
```diff
@@ -53,7 +53,24 @@ def find_splits(ct_df: pd.DataFrame):
     return ct_df
 
 
-def split_tasks_into_separate_files(path: str, result_name_suffix='tasks'):
+# to find start index for each group of rows with the same task
```
nbirillo: Please, capitalize the first letter

src/main/util/consts.py
elena-lyulina: maybe it's better to use TASK.__members__.items() here also?

nbirillo: `TASK.members` returns dictionary and each key has string type in this case

PR #20, opened 2020-03-23
Path finder

nbirillo:
Create structure and implement a base version of algo


src/main/solution_space/path_finder.py
```diff
@@ -0,0 +1,166 @@
...
+            if diffs <= user_diffs_to_goal:
+                candidates.append(vertex)
+        return self.__choose_best_vertex(user_code, user, vertices)
+
+    # Choose the best way to go to the goal
+    # For example, if we have a good way through the graph, we should advise it,
+    # but if don't we would advise going directly to the goal
+    @staticmethod
+    def __go_through_graph(user_code: Code, graph_vertex: Vertex, goal: Vertex,
+                           median_of_diff_number_to_goal: int) -> bool:
```
elena-lyulina: maybe we should pass the graph as argument and get median_of_diff_number_to_goal inside the method? We don't use it anywhere else except there + our algorithm can change, but we probably need a graph in this method anyway

src/main/solution_space/solution_graph.py
```diff
@@ -228,3 +231,32 @@ def add_code_info_chain(self, code_info_chain: Optional[List[Tuple[Code, CodeInf
...
+    def update_ages_and_experiences(code_info_list: List[CodeInfo],
+                                    ages: List[int],
+                                    experiences: List[Any]) -> None:
+        for code_info in code_info_list:
+            if not code_info.user.profile.age != DEFAULT_VALUES.AGE.value:
+                ages.append(code_info.user.profile.age)
+            if not code_info.user.profile.experience != DEFAULT_VALUES.EXPERIENCE:
+                experiences.append(code_info.user.profile.experience)
+
+    def calculate_median_of_profile_info(self) -> Tuple[int, Any]:
```
elena-lyulina: I'm not sure that we need a median for a profile in this pr, because we need to have a small and simple implementation of path finder. Maybe it would be better to move them into separate branch and leave for better times?

src/main/solution_space/path_finder.py
```diff
@@ -0,0 +1,166 @@
...
+    @property
+    def users_count(self) -> int:
+        return self._users_count
+
+    def __init_profile(self, user: User) -> Profile:
+        # Todo: add user handling
+        ages, experiences = [], []
+        SolutionGraph.update_ages_and_experiences(self._vertex.code_info_list, ages, experiences)
+        return Profile(calculate_safety_median(ages, default_value=DEFAULT_VALUES.AGE.value),
+                       calculate_median_for_objects(experiences, default_value=DEFAULT_VALUES.EXPERIENCE))
```
elena-lyulina: Why don't you call calculate_median_of_profile_info(self)? However, i think we shouldn't init Ptofil with median right now, because we want to have as simple as possible version of PathFinder. Let's init it with None and leave medians for other pr?

src/main/solution_space/solution_graph.py
```diff
@@ -228,3 +231,32 @@ def add_code_info_chain(self, code_info_chain: Optional[List[Tuple[Code, CodeInf
                 prev_vertex.add_child(next_vertex)
                 prev_vertex = next_vertex
         log.info(f'Finish adding code-user chain')
+
+    @staticmethod
+    def update_ages_and_experiences(code_info_list: List[CodeInfo],
```
elena-lyulina: I don't fully understand the sense of this method. All two times that it's called, you pass en empty lists for ages and experiences, so why don't you want to get ages and experiences directly, without passing any lists? Also, if these lists were not empty, they would contain some incorrect (or doubled) elements, because you don't check what they contain

src/main/util/statistics_util.py
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2020 Anastasiia Birillo, Elena Lyulina
+
+from statistics import median
+from collections import Counter
+from typing import List, Any, Optional, Union
+
+
+def calculate_safety_median(values: List[int],
```
elena-lyulina: Safety is a noun, so let's rename it to calculate_median_safely

src/main/solution_space/path_finder.py
```diff
@@ -0,0 +1,166 @@
...
+        vertices = self._graph.get_traversal()
+        vertices.remove(self._graph.start_vertex)
+        for vertex in vertices:
+            # We don't want to add to result the same vertex
+            if are_asts_equal(user_code.ast, vertex.code.ast):
+                continue
+            diffs = get_diffs_number(vertex.code.file_with_code, goal.code.file_with_code)
+            if diffs <= user_diffs_to_goal:
+                candidates.append(vertex)
+        return self.__choose_best_vertex(user_code, user, vertices)
```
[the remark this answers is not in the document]
nbirillo: Thanks! It is my mistake

src/main/solution_space/path_finder.py
```diff
@@ -0,0 +1,161 @@
...
+    @staticmethod
+    def __apply_minimal_actions_number(user_code: Code, dst_code: Code) -> Code:
+        str_code = apply_diffs(user_code.file_with_code, dst_code.file_with_code)
+        ast = get_canonicalized_form(str_code)
+        # Todo: get rate
+        return Code(ast)
+
+
+# Todo: rename it???
+class Candidate:
```
elena-lyulina: rename it to MeasuredVertex, please

src/main/solution_space/path_finder.py
```diff
@@ -0,0 +1,161 @@
...
+        if not isinstance(o, Candidate):
+            log.error(f'The object {o} is not {self.__class__} class')
+            raise ValueError(f'The object {o} is not {self.__class__} class')
+        if self._distance < o.distance:
+            return True
+        # Todo: use profile info
+        # if self.profile.experience < o.profile.experience:
+        #     return True
+        # if self.profile.age < o.profile.age:
+        #     return True
```
elena-lyulina: Let's remove these comments since they are related to another branch

src/main/util/consts.py
```diff
@@ -72,6 +71,14 @@ def is_equal(self, value) -> bool:
             return isnat(value)
         return equal(self.value, value)
 
+    def __lt__(self, o: object) -> bool:
```
elena-lyulina: let's move it to the separate branch also

src/main/util/consts.py
```diff
@@ -81,6 +88,15 @@ class EXPERIENCE(Enum):
     FROM_FOUR_TO_SIX_YEARS = 'FROM_FOUR_TO_SIX_YEARS'
     MORE_THAN_SIX = 'MORE_THAN_SIX'
 
+    def __lt__(self, o: object) -> bool:
```
elena-lyulina: and this :)

src/test/util/const/experience_tests.py
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) 2020 Anastasiia Birillo, Elena Lyulina
+
+import logging
+import unittest
+
+from src.main.util.consts import LOGGER_TEST_FILE, LOGGER_FORMAT, EXPERIENCE, DEFAULT_VALUES
+
+
+class TestExperienceMethods(unittest.TestCase):
```
elena-lyulina: I think, that also should be in a separate branch

src/test/solution_space/path_finder/choose_best_vertex_tests.py
```diff
@@ -0,0 +1,53 @@
...
+        sg.connect_to_start_vertex(v)
+        sg.connect_to_end_vertex(v)
+    return sg
+
+
+SG = init_graph()
+PF = PathFinder(SG)
+
+
+def run_empty_list_test() -> Optional[Vertex]:
```
elena-lyulina: Looking at the realization, you don't run any tests, you choose the best vertex. Please, rename it to be consistent with method behavior.

src/main/solution_space/consts.py
```diff
@@ -15,5 +15,9 @@ class VERTEX_TYPE(Enum):
 
 FOLDER_WITH_CODE_FILES = ROOT_DIR + '/../../resources/solution_space'
 FOLDER_WITH_CODE_FILES_FOR_TESTS = ROOT_DIR + '/../../resources/test_data/solution_space'
+EMPTY_CODE_FILE = ROOT_DIR + '/../../resources/solution_space/empty_code.py'
 GRAPH_FOLDER_PREFIX = 'graph_'
 FILE_PREFIX = 'code_'
+
+DIFFS_PERCENT_GO_DIRECTLY = 0.2
```
elena-lyulina: DIFFS_PERCENT_GO_DIRECTLY -> DIFFS_PERCENT_TO_GO_DIRECTLY

src/main/solution_space/consts.py
```diff
@@ -15,5 +15,9 @@ class VERTEX_TYPE(Enum):
...
+TO_GRAPH_DISTANCE_THRESHOLD = 2
```
elena-lyulina: TO_GRAPH_DISTANCE_THRESHOLD -> DISTANCE_TO_GRAPH_THRESHOLD

src/main/solution_space/path_finder.py
```diff
@@ -75,25 +72,26 @@ def __find_closest_vertex_with_path(self, user_code: Code, user: User, goal: Ver
...
-        return user_diffs_to_graph / user_diffs_to_goal > TO_GRAPH_DISTANCE_THRESHOLD
+    def __is_far_from_graph(diffs_from_user_to_goal: int, diffs_from_user_to_graph_vertex: int) -> bool:
+        return diffs_from_user_to_graph_vertex / diffs_from_user_to_goal > DISTANCE_TO_GRAPH_THRESHOLD
 
     @staticmethod
-    def __is_almost_goal(empty_diffs_to_goal: int, user_diffs_to_goal: int) -> bool:
-        return abs(empty_diffs_to_goal - user_diffs_to_goal) <= empty_diffs_to_goal * DIFFS_PERCENT_GO_DIRECTLY
+    def __is_most_of_path_is_done(diffs_from_empty_to_goal: int, diffs_from_user_to_goal: int) -> bool:
+        return diffs_from_user_to_goal \
+               <= (diffs_from_empty_to_goal + diffs_from_user_to_goal) * DIFFS_PERCENT_TO_GO_DIRECTLY
```
elena-lyulina: Sorry, but I think it's something wrong with this formula again.
shouldn't we sum diffs_from_empty_to_user and diffs_from_user_to_goal?

PR #36, opened 2020-04-28
Solution space serializer

nbirillo:


src/main/solution_space/data_classes.py
```diff
@@ -38,7 +38,9 @@ def is_empty(self) -> bool:
                and DEFAULT_VALUE.EVENT_DATA.is_equal(self._event_data)
 
     def __str__(self) -> str:
-        return f'Timestamp: {self._timestamp}, event_type: {self._event_type}, event_data: {self._event_data}'
+        return f'\n\n________________ATI ITEM START________________\n\n' \
```
elena-lyulina, line 41: I'm a little bit confused by these '_______________' symbols. It can be a little tedious to write them every time. Maybe we need a method (somewhere in util?) that wraps string into them? Therefore we can simply change the format if we want. 
Also, sometimes we need _ _ str _ _ method not only for graph serialization, so I'm not sure at all that adding additional symbols is a good  idea. Maybe we need an additional pretty_print function that will print our data in this format? So we can leave _ _ str _ _ just returning all fields without any formatting.

PR #40, opened 2020-05-04
Boost algo

nbirillo:
Plot statistics about solution graph


src/main/plots/solution_graph_statistics_plots.py
```diff
@@ -0,0 +1,107 @@
...
+            title_text=get_readable_key(column.value),
+            categoryorder=x_category_order.value
+        ),
+        plot_bgcolor=STATISTICS_COLORS.BAR_CHART_BG.value
+    )
+    fig.update_yaxes(automargin=True)
+    save_plot(fig, path, PLOT_TYPE.BAR, plot_name, format, auto_open)
+
+
+def plot_general_statistics(solution_graph: SolutionGraph, path: str = DEFAULT_PATH_FOR_PLOTS,
```
elena-lyulina, line 47: 'General statistics' is a little bit unclear because it can mean anything. How about 'plot_all_node_numbers_freq'?

src/main/plots/solution_graph_statistics_plots.py
```diff
@@ -0,0 +1,107 @@
...
+                            format: consts.EXTENSION = consts.EXTENSION.HTML, auto_open: bool = False,
+                            x_category_order: PLOTTY_CATEGORY_ORDER = PLOTTY_CATEGORY_ORDER.TOTAL_ASCENDING) -> None:
+    canon_trees_statistics, anon_trees_statistics = get_general_solution_graph_statistics(solution_graph)
+    labels = get_labels_for_freq_plots(STATISTICS_KEY.NODES_NUMBER)
+    __plot_general_statistics(canon_trees_statistics, title='Vertices number for canon trees', path=path,
```
elena-lyulina, line 52: Let's use plot_freq_chart from plotty_util instead of this

src/main/plots/solution_graph_statistics_plots.py
```diff
@@ -0,0 +1,107 @@
...
+            ),
+            showlegend=False,
+            marker_symbol='x',
+            hoverinfo='skip'
+        )
+    )
+    return fig
+
+
+def plot_statistics_for_vertex(vertex_id: int, labels: Dict[str, str], anon_nodes_numbers: List[int],
```
elena-lyulina, line 80: How about 'plot_node_numbers_freq_for_vertex'? Or maybe it's too long, but I'm afraid that otherwise, we can forget about what does this function plot after some time

src/main/plots/solution_graph_statistics_plots.py
```diff
@@ -0,0 +1,107 @@
...
+    fig = get_freq_bar_chart(statistics_df, title, column, labels,
+                             x_category_order=PLOTTY_CATEGORY_ORDER.TRACE,
+                             to_update_layout=False)
+    __add_canon_nodes_numbers(fig, canon_nodes_number, freq_dict)
+    # Todo: fix x_category_order, because now order is not always ascending
+    fig = update_layout(fig, column, PLOTTY_CATEGORY_ORDER.TRACE)
+    save_plot(fig, path, PLOT_TYPE.BAR, f'vertex_{vertex_id}', format, auto_open)
+
+
+def __get_tittle_for_each_vertex_plots(vertex_id_: int, column: STATISTICS_KEY = STATISTICS_KEY.NODES_NUMBER) -> str:
```
elena-lyulina, line 97: It's a little unclear what do you mean by 'for each vertex'. In the function below you use it when you iterating over all vertices, but in this function, you just get a title for one vertex, which id is passed as an argument. Let's rename it to 'get_title_for_vertex_plot'?

src/main/plots/solution_graph_statistics_plots.py
```diff
@@ -0,0 +1,107 @@
...
+    return f'{get_readable_key(column.value)} distribution for anon trees for vertex with id {vertex_id_}'
+
+
+def plot_statistics_for_each_vertex(solution_graph: SolutionGraph, path: str = DEFAULT_PATH_FOR_PLOTS,
```
elena-lyulina, line 101: Again, let's be more specific about which statistics do we plot

src/main/plots/util/plotly_util.py
```diff
@@ -1,20 +1,68 @@
...
+                       to_update_layout: bool = True) -> go.Figure:
+    fig = px.bar(statistics_df, x=column.value, y=STATISTICS_FREQ, title=title, labels=labels,
+                 hover_data=[column.value, STATISTICS_FREQ])
+    if to_update_layout:
+        fig = update_layout(fig, column, x_category_order)
+    fig.update_yaxes(automargin=True)
+    return fig
+
+
+def plot_freq_chart(statistics_df: pd.DataFrame, title: str, path: str, column: STATISTICS_KEY,
```
elena-lyulina, line 62: Maybe it's better to include 'save' in the name? It's pretty important to mention it because it's the only difference between the function above.

src/main/statistics_gathering/solution_graph_statistics.py
```diff
@@ -0,0 +1,43 @@
...
+
+from collections import defaultdict
+from typing import Tuple, Dict, Union, List
+
+from src.main.canonicalization.consts import TREE_TYPE
+from src.main.solution_space.solution_graph import SolutionGraph
+from src.main.canonicalization.ast_tools import get_vertices_number_in_ast
+
+
+def get_general_solution_graph_statistics(solution_graph: SolutionGraph) -> Tuple[Dict[int, int], Dict[int, int]]:
```
elena-lyulina, line 11: Again, let's be more specific about statistics. We can have a different statistic for our graph in the future, so we should always describe it in detail.

src/main/statistics_gathering/solution_graph_statistics.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def __get_default_dict_for_vertex() -> dict:
+    return {
+        TREE_TYPE.CANON: 0,
+        TREE_TYPE.ANON: []
+    }
+
+
+def get_statistics_for_each_vertex(solution_graph: SolutionGraph) -> dict:
```
elena-lyulina, line 33: Again, add statistics detail to the name. Mention 'node_numbers', please.

elena-lyulina:
Good job :)

PR #41, opened 2020-05-05
Add dict in IdCounter for getting items by id

nbirillo:


src/main/util/default_dict_util.py
```diff
@@ -0,0 +1,19 @@
...
+
+def get_none() -> None:
+    return None
+
+
+def get_empty_list() -> list:
+    return []
+
+
+def get_default_dict_with_default_dict_with_none() -> dict:
```
elena-lyulina, line 18: let's reduce the number of 'default_dict' in this name :)

src/main/util/helper_classes/id_counter.py
```diff
@@ -1,15 +1,25 @@
...
+from src.main.util.consts import LOGGER_NAME
+from src.main.util.log_util import log_and_raise_error
+from src.main.util.default_dict_util import get_default_dict_with_default_dict_with_none, get_none
+
+log = logging.getLogger(LOGGER_NAME)
+
 
 class IdCounter:
     _instances: Dict[str, int] = defaultdict(int)
+    _id_item_dict_by_class: Dict[str, Dict[int, Any]] = defaultdict(get_default_dict_with_default_dict_with_none)
```
elena-lyulina, line 16: I think since we store items in this dict, their type is not Any, but Type[IdCounter], what do you think?

elena-lyulina:
Cool, thanks!

PR #42, opened 2020-05-05
Solution space/path finder improving

elena-lyulina:
This PR includes a lot of small improvements such as:
1. Creating empty vertex in the graph
2. Add ability to find rate during Code creation
3. Change traversal a little to be able to include/not include start and end vertices
4. Change test_system so now it's possible not to print equal docs
5. Add multithreading to distance and remove it from graph, since we don't use it anymore


src/main/solution_space/distance.py
```diff
@@ -1,44 +1,88 @@
 # Copyright (c) 2020 Anastasiia Birillo, Elena Lyulina
```
nbirillo, line 1: Please, write **Todo comment** with my comments in the file, we will fix it in the future because now we don't use distance

src/main/solution_space/path_finder/path_finder.py
```diff
@@ -4,13 +4,10 @@
 
 import logging
 from abc import ABCMeta, abstractmethod
+from typing import Type, Optional
```
nbirillo, line 7: Import's order

src/main/solution_space/path_finder_test_system.py
```diff
@@ -69,18 +71,25 @@ def __init__(self, test_inputs: List[TestInput], serialized_graph_path: str = TE
...
+        methods_dict, methods_names = self.__get_methods_dict_and_names(classes, methods_to_keep)
+        table = PrettyTable(title=title)
+        table.add_column('class_name', [c.__name__ for c in classes])
+        for name in methods_names:
+            methods = [methods_dict[c].get(name) for c in classes]
+            docs = [self._no_method_sign if m is None else self.__format_doc_str(m.__doc__) for m in methods]
+            if not self._add_same_docs and len(set(docs)) == 1:
+                continue
+            table.add_column(name, docs)
+        # for c in classes:
```
nbirillo, line 85: Should we delete the comment?

src/main/util/consts.py
```diff
@@ -192,7 +192,7 @@ class SPLIT(Enum):
     TASKS = 'tasks'
 
 
-LOGGER_FILE = ROOT_DIR + '../../../../logs.log'
+LOGGER_FILE = ROOT_DIR + '../../../../logs_1.log'
```
nbirillo, line 195: Why did you rename it?

src/test/solution_space/distance/multithreaded_filling_dist_test.py
```diff
@@ -0,0 +1,49 @@
+# Copyright (c) 2020 Anastasiia Birillo, Elena Lyulina
+import logging
```
nbirillo, line 2: Please, add an empty line

src/main/solution_space/distance.py
```diff
@@ -1,44 +1,88 @@
...
-            return False
-
         if new_item in self._dist.keys():
             log.info('This item already exists')
             return False
+
+        # If we don't store dist, we don't fill self._dist[new_item],
+        # but we add an empty dict to be able to fill self._dist later
         self._dist[new_item] = {}
+        if not self._to_store_dist:
```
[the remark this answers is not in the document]
elena-lyulina: No, we shouldn't. We need to store an empty dict for each vertex because when we will fill dist_matrix we should have all vertices as keys. 
If we swap these lines, we won't be able to fill dist matrix.
I wrote a comment about it right above line 81.

src/main/solution_space/path_finder_test_system.py
```diff
@@ -193,4 +200,16 @@ def __get_all_subclasses(cls, clazz: Type[Class]) -> List[Type[Class]]:
...
+        not_skipped_subclasses = subclasses.copy()
+        for c in subclasses:
+            if hasattr(c, 'is_skipped') and c.is_skipped:
+                not_skipped_subclasses.remove(c)
+                TestSystem.__print_output(f'{c.__name__} is skipped, reason: {c.skipped_reason}')
+        return not_skipped_subclasses
+
+    # Todo: add ability to print output to file?
+    @staticmethod
+    def __print_output(output: Any) -> None:
```
nbirillo, line 213: Please, use `Optional[Any]` for the `output ` param

PR #44, opened 2020-05-06
Add class for anon tree

elena-lyulina:


src/main/solution_space/serialized_code.py
```diff
@@ -5,8 +5,9 @@
 import os
 import ast
 import logging
-from typing import List, Dict, Callable, Optional
+from typing import List, Dict, Callable, Optional, Set
 
+from src.main.solution_space.data_classes import CodeInfo, User
```
nbirillo, line 10: Imports order

src/main/solution_space/serialized_code.py
```diff
@@ -21,6 +22,31 @@
 log = logging.getLogger(consts.LOGGER_NAME)
 
 
+class AnonTree(IdCounter):
```
nbirillo, line 25: Please, add **PrettyString** too for possibilities to get a pretty string

src/main/solution_space/serialized_code.py
```diff
@@ -102,18 +129,21 @@ def language(self) -> consts.LANGUAGE:
...
+        if found_anon_tree:
+            found_anon_tree.add_code_info(code_info)
             return None
 
-        self._anon_trees.append(anon_tree)
-        index_of_tree = self._anon_trees.index(anon_tree)
-        return self.__create_file_for_tree(anon_tree, f'{TREE_TYPE.ANON.value}_{index_of_tree}')
+        new_anon_tree = AnonTree(anon_tree, code_info)
+        self._anon_trees.append(new_anon_tree)
+        index_of_tree = self._anon_trees.index(new_anon_tree)
```
nbirillo, line 140: We use simple list and the append function always add items into the end of the list. Maybe should we replace `self._anon_trees.index(new_anon_tree)` to `len(self._anon_trees)`? It would be faster than your implementation

src/test/solution_space/solution_graph/graph_test.py
```diff
@@ -101,7 +102,11 @@ def create_code_info_chain() -> (List[Tuple[Code, CodeInfo]], List[str]):
 
 def get_vertex_structure(vertex: Vertex) -> VertexStructure:
     source = get_code_from_tree(vertex.serialized_code.canon_tree).strip('\n') if vertex.serialized_code else None
-    return {VERTEX_STRUCTURE.SOURCE: source, VERTEX_STRUCTURE.CODE_INFO_LIST_LEN: len(vertex.code_info_list)}
+    if vertex.serialized_code is None:
```
nbirillo, line 105: Maybe is it better `code_info_list_len = 0 if vertex.serialized_code is None else sum([len(a_t.code_info_list) for a_t in vertex.serialized_code.anon_trees])`?

src/main/solution_space/serialized_code.py
```diff
@@ -102,18 +129,21 @@ def language(self) -> consts.LANGUAGE:
...
+        return self.__create_file_for_tree(new_anon_tree.tree, f'{TREE_TYPE.ANON.value}_{index_of_tree}')
```
nbirillo, line 141: Should we add the tree to the anon number vertices duct?

PR #48, opened 2020-05-07
Path finder v 3 new

nbirillo:
New algo:
        1. Find the same tree SAME_TREE in the graph and get the best tree from next trees (__find_same_tree_in_graph)
        2. If SAME_TREE is not None, return SAME_TREE
        2. Find the closest tree CLOSEST_TREE in graph (__find_closest_tree with graph.canon_trees_nodes_number)
        3. If __is_most_of_path_is_done, return CLOSEST_TREE
        4. Find the closest goal CLOSEST_GOAL in graph (__find_closest_goal_tree)
        5. Choose between CLOSEST_TREE and CLOSEST_GOAL using __go_through_graph


src/main/solution_space/path_finder/path_finder_v_3.py
```diff
@@ -41,63 +51,58 @@ def find_next_anon_tree(self, user_anon_tree: AnonTree, user_canon_tree: ast.AST
...
-        canon_nodes_number = get_nodes_number_in_ast(user_canon_tree)
-        anon_nodes_number = user_anon_tree.nodes_number
+    @staticmethod
+    def __get_nodes_number_dict(nodes_number: List[Any], filter_foo: Callable = lambda x: x) -> Dict[int, list]:
+        return {k: list(v) for k, v in groupby(nodes_number, filter_foo)}
+
+    # Note: we have to remove the 'user_code' from the set
+    def __find_closest_tree(self, user_anon_tree: AnonTree, user_canon_nodes_number: int,
+                            user_anon_tree_nodes_number: int,
+                            canon_trees_nodes_number: Dict[int, list]) -> Optional[AnonTree]:
```
elena-lyulina, line 75: Let's call an argument `canon_nodes_numbers_dict` instead of  `canon_trees_nodes_number` because now it looks like the other argument, `user_canon_nodes_number`

src/main/solution_space/path_finder/path_finder_v_3.py
```diff
@@ -41,63 +51,58 @@ def find_next_anon_tree(self, user_anon_tree: AnonTree, user_canon_tree: ast.AST
...
+        """
+        1. Consider each vertex with similar nodes number as candidate (chose at least TOP_N_CANON candidates)
```
elena-lyulina, line 77: Please, add the second item: 
2. Choose at least TOP_N_ANON anon trees from canon candidates and run __choose_best_anon_tree

src/main/solution_space/path_finder/path_finder_v_3.py
```diff
@@ -41,63 +51,58 @@ def find_next_anon_tree(self, user_anon_tree: AnonTree, user_canon_tree: ast.AST
...
         # Get vertices ids with canon trees, which have nodes number similar to user canon_nodes_number
-        vertices_ids = self.__get_top_n_candidates(CANON_TOP_N, canon_nodes_number, self._graph.canon_trees_nodes_number)
+        vertices_ids = self.__get_top_n_candidates(CANON_TOP_N, user_canon_nodes_number, canon_trees_nodes_number)
         vertices: List[Vertex] = [Vertex.get_item_by_id(id) for id in vertices_ids]
 
         anon_trees = sum([v.serialized_code.anon_trees for v in vertices], [])
-        anon_nodes_numbers_dict = {k: list(v) for k, v in groupby(anon_trees, lambda a_t: a_t.nodes_number)}
-        anon_candidates = self.__get_top_n_candidates(ANON_TOP_N, anon_nodes_number, anon_nodes_numbers_dict)
-
+        anon_nodes_numbers_dict = self.__class__.__get_nodes_number_dict(anon_trees, lambda a_t: a_t.nodes_number)
```
elena-lyulina, line 85: You can call classmethods without calling `__class__` because `self` is an instance of class. So just call self.__get_nodes_number_dict

src/main/solution_space/path_finder/path_finder_v_3.py
```diff
@@ -41,63 +51,58 @@ def find_next_anon_tree(self, user_anon_tree: AnonTree, user_canon_tree: ast.AST
...
 
-    def __find_closest_tree_from_equal_tree(self, user_anon_tree: AnonTree, equal_anon_tree: AnonTree) -> AnonTree:
-        log.info('Found the same anon tree')
-        # Todo: what to do if there is the exact vertex? Find a path to the goal? Now just returning it
-        return equal_anon_tree
+    def __is_most_of_path_is_done(self, closest_tree: AnonTree) -> bool:
+        """
+        1. Use only nodes number info. Use median for goals nodes number
+        """
+        return closest_tree.nodes_number <= \
```
elena-lyulina, line 93: Shouldn't we use user_anon_tree here instead of closesr_tree? Also, let's call this method '__is_close_to_goals'  instead of '__is_most_of_path_is_done' because now we don't know our path (before we knew that our path is the path to the closest goal, but now we don't have the closest goal)

src/main/solution_space/path_finder/path_finder_v_3.py
```diff
@@ -41,63 +51,58 @@ def find_next_anon_tree(self, user_anon_tree: AnonTree, user_canon_tree: ast.AST
...
+               self.graph.get_median_goals_nodes_numbers() * NODES_NUMBER_PERCENT_TO_GO_DIRECTLY
+
+    def __find_closest_goal_tree(self, user_anon_tree: AnonTree, user_canon_nodes_number: int,
+                                 user_anon_tree_nodes_number: int, ) -> AnonTree:
+        """
+        1. Get list of all goals
+        2. Chose at least TOP_N_CANON candidates
+        2. Find the closest using __choose_best_vertex()
+        """
+        canon_nodes_numbers_dict = self.__class__.__get_nodes_number_dict(self.graph.goals_nodes_number)
```
elena-lyulina, line 103: You can call classmethods without calling `__class__` because `self` is an instance of class. So just call self.__get_nodes_number_dict

src/main/solution_space/path_finder/path_finder_v_3.py
```diff
@@ -134,14 +139,23 @@ def __get_top_n_candidates(top_n: int, nodes_number: int, nodes_numbers_dict: Di
...
+        candidates.sort()
+        log.info(f'The best vertex id is {candidates[0].candidate_tree.id}')
+        return candidates[0].candidate_tree
+
     @staticmethod
     def __is_far_from_graph(diffs_from_user_to_goal: int, diffs_from_user_to_graph_vertex: int) -> bool:
         return diffs_from_user_to_graph_vertex / diffs_from_user_to_goal > DISTANCE_TO_GRAPH_THRESHOLD
 
-    @staticmethod
-    def __is_most_of_path_is_done(diffs_from_empty_to_goal: int, diffs_from_user_to_goal: int) -> bool:
```
elena-lyulina, line 142: Maybe we should keep this method? We just checked the number of nodes, but maybe we still want to check distance?

src/main/solution_space/serialized_code.py
```diff
@@ -105,6 +110,12 @@ def tree(self) -> ast.AST:
     def code_info_list(self) -> List[CodeInfo]:
         return self._code_info_list
 
+    def add_next_anon_tree(self, next_anon_tree: 'AnonTree') -> bool:
+        if next_anon_tree.id in self._next_anon_trees_ids:
+            return False
+        self._next_anon_trees_ids.append(next_anon_tree.id)
+        return True
+
```
elena-lyulina, line 118: Maybe we should also check that next_anon_tree.id is not self.id? It can happen, but we don't want to pass the same tree in choose_best_anon_tree, so we should be sure that it's not in the next_anon_trees_ids

src/main/solution_space/solution_graph.py
```diff
@@ -18,7 +19,7 @@
 from src.main.solution_space import consts as solution_space_consts
 from src.main.util.file_util import remove_directory, create_directory
 from src.main.util.consts import LOGGER_NAME, TASK, LANGUAGE, TEST_RESULT
-from src.main.canonicalization.canonicalization import are_asts_equal, get_code_from_tree
+from src.main.canonicalization.canonicalization import are_asts_equal, get_code_from_tree, get_nodes_number_in_ast
 from src.main.solution_space.consts import GRAPH_FOLDER_PREFIX, SOLUTION_SPACE_FOLDER, FILE_PREFIX
```
elena-lyulina, line 23: imports order

src/main/solution_space/solution_graph.py
```diff
@@ -114,6 +121,9 @@ def language(self) -> LANGUAGE:
     def empty_vertex(self) -> Vertex:
         return self._empty_vertex
 
+    def get_median_goals_nodes_numbers(self):
+        return median(self._goals_nodes_number)
```
elena-lyulina, line 125: Please, add try-except block in case of the empty list

src/main/solution_space/serialized_code.py
```diff
@@ -189,6 +200,11 @@ def language(self) -> consts.LANGUAGE:
     def is_full(self) -> bool:
         return self._rate == consts.TEST_RESULT.FULL_SOLUTION.value
 
+    def get_last_anon_tree(self) -> Optional[AnonTree]:
+        if not self.anon_trees:
+            return None
+        return self.anon_trees[len(self.anon_trees) - 1]
```
elena-lyulina, line 206: You can use self.anon_trees[-1] to get the last tree

src/main/solution_space/solution_graph.py
```diff
@@ -188,6 +199,9 @@ def add_code_info_chain(self, code_info_chain: Optional[List[Tuple[Code, CodeInf
             for next_code, next_code_info in code_info_chain[1:]:
                 next_vertex = self.find_or_create_vertex(next_code, next_code_info)
                 prev_vertex.add_child(next_vertex)
+                next_anon_tree = next_vertex.serialized_code.get_last_anon_tree()
+                if next_anon_tree:
+                    prev_vertex.serialized_code.get_last_anon_tree().add_next_anon_tree(next_anon_tree)
```
elena-lyulina, line 204: Sorry, but now I'm not sure it works :(
Maybe we should have a variable 
`next_anon_tree = next_vertex.serialized_code.find_anon_tree(next_code.anon_tree)`?

PR #49, opened 2020-05-07
Measured tree distance

elena-lyulina:


src/main/solution_space/measured_tree/measured_tree_v_1.py
```diff
@@ -0,0 +1,53 @@
...
+        5. (if possible) abs difference between age, weight: {4}
+        6. (if possible) abs difference between exp, weight: {5}
+        """
+
+        distance = self._diffs_w * self._diffs_number\
+                   - self._users_w * self.users_count\
+                   + self._rate_w * (self.user_tree.rate - self.candidate_tree.rate)\
+                   + self._rollback_w * self.rollback_probability
+
+        if not (self.user_tree.has_empty_age() or self.candidate_tree.has_empty_age()):
```
nbirillo, line 37: Maybe can we create a static method in the AnonTree class with the following signature:
`def has_not_empty_value(value, *args)`, where **value** can be `AGE` or `EXPERIENCE` and ***args** is the list with trees
?

I think it is better to use the general function because we can add an additional param in the future. Also, it is to hard to read the long-expressions

src/main/solution_space/path_finder_test_system.py
```diff
@@ -17,6 +17,7 @@
 
 from prettytable import PrettyTable, ALL
 
+from src.main.preprocessing.int_experience_adding import convert_to_int_experience
```
nbirillo, line 20: Imports order

src/main/solution_space/path_finder_test_system.py
```diff
@@ -130,10 +138,11 @@ def __get_all_possible_path_finders(self) -> List[IPathFinder]:
         return path_finders
 
     def __create_user_trees(self, test_input: TestInput) -> Tuple[AnonTree, ast.AST]:
-        code_info = CodeInfo(User(Profile(test_input[TEST_INPUT.AGE], test_input[TEST_INPUT.EXPERIENCE])))
+        code_info = CodeInfo(User(Profile(test_input[TEST_INPUT.AGE], test_input[TEST_INPUT.INT_EXPERIENCE])))
         code = Code.from_source(test_input[TEST_INPUT.SOURCE_CODE], rate=None, task=self._graph.task)
         serialized_code = SerializedCode(code, code_info, self._graph.graph_directory, self._graph.file_prefix)
         anon_tree = serialized_code.anon_trees[0]
+        anon_tree.find_medians()
```
nbirillo, line 145: You often use find_medians after getting an anon tree. Maybe should we have a method **get_anon_tree_by_index** with bool param **to_find_medians**?

src/main/solution_space/serialized_code.py
```diff
@@ -5,10 +5,12 @@
 import os
 import ast
 import logging
+from statistics import median, StatisticsError
 from typing import List, Callable, Optional, Set
 
+from src.main.solution_space.consts import EMPTY_MEDIAN
```
nbirillo, line 11: Imports order

src/main/solution_space/path_finder_test_system.py
elena-lyulina: Hmm, it's only when we create a new AnonTree for user (and it happens twice: here and in Hint class). I guess we should have a method that does all these steps:
1. creates Code from source
2. creates SerializedCode
3. gets the first anon_tree in serialized_code
5. finds anon medians

I think, that beyond these steps we don't need to get anon tree by index and call to_find_medians

PR #52, opened 2020-05-08
Add serialization with restoring IdCounter variables

elena-lyulina:


src/main/main.py
```diff
@@ -7,6 +7,7 @@
 import logging
 from datetime import datetime
 
+import dill
```
nbirillo, line 10: Why did you add it?

src/main/solution_space/solution_space_serializer.py
```diff
@@ -18,25 +19,31 @@ class SolutionSpaceSerializer:
...
+        objects_to_serialize = (graph, IdCounter._instances, IdCounter._id_item_dict_by_class)
+        serialize_data_and_write_to_file(graph_path, objects_to_serialize)
+        log.info(f'Graph with id {graph.id} was serialized successfully. File path is {graph_path}')
+        return graph_path
 
     @staticmethod
     def deserialize(serialized_graph_path: str,
                     new_path_for_graph: Optional[str] = None,
                     to_delete_old_graph_directory: bool = True) -> Optional[SolutionGraph]:
+        # instances_path = add_suffix_to_file(serialized_graph_path, 'instances')
```
nbirillo, line 33: Should we delete it?

src/main/solution_space/solution_space_serializer.py
```diff
@@ -18,25 +19,31 @@ class SolutionSpaceSerializer:
     # Returns file path with serialized graph
     @staticmethod
     def serialize(graph: SolutionGraph, serialized_file_prefix: str = 'serialized_graph') -> str:
-        path = os.path.join(SERIALIZED_GRAPH_PATH, f'graph_{graph.id}_{serialized_file_prefix}{EXTENSION.PICKLE.value}')
-        serialize_data_and_write_to_file(path, graph)
-        log.info(f'Graph with id {graph.id} was serialized successfully. File path is {path}')
-        return path
+        graph_name = f'graph_{graph.id}_{serialized_file_prefix}'
+        graph_path = os.path.join(SERIALIZED_GRAPH_PATH, f'{graph_name}{EXTENSION.PICKLE.value}')
+        objects_to_serialize = (graph, IdCounter._instances, IdCounter._id_item_dict_by_class)
```
nbirillo, line 24: I am not sure that we should serialize all instances. For example, before serializing process we have 2 users and the last Id will be 2, but we can stop the program and run the program with the deserializing process. And we will have the last Id = 2. Is it the correct case?

src/main/util/helper_classes/id_counter.py
```diff
@@ -36,3 +35,11 @@ def get_item_by_id(cls: 'IdCounter', id: int) -> 'IdCounter':
         if item is None:
             log_and_raise_error(f'Item with id {id} does not exist in the class {cls.__name__}', log)
         return item
+
+    def set_item_by_id(self) -> None:
```
nbirillo, line 39: You use it only in the tests. Maybe should we not have the method here? And implement it only in the tests

src/test/solution_space/solution_space_serializer/solution_space_serializer_test.py
```diff
@@ -46,3 +49,9 @@ def test_graph_representation(self, param_solution_space_serializer_test: Callab
         deserialized_graph = SolutionSpaceSerializer.deserialize(serialized_path)
         assert are_graph_folder_structures_equal(graph.graph_directory, deserialized_graph.graph_directory)
         assert graph.get_pretty_string() == deserialized_graph.get_pretty_string()
+
+    def test_id_counter_serialization(self) -> None:
```
nbirillo, line 53: It is a strange method. You don't use assert or others. Maybe is it better to use set_item_by_id method here with assert id's for example?

PR #53, opened 2020-05-08
Add new task

nbirillo:
I am not sure the way to save anon trees is good, but I hope you can help me with it


src/main/solution_space/path_finder/path_finder_v_3.py
```diff
@@ -73,7 +73,7 @@ def __find_same_tree_in_graph(self, user_anon_tree: AnonTree, user_canon_tree: a
             if graph_anon_tree:
                 next_anon_trees = [AnonTree.get_item_by_id(id) for id in graph_anon_tree.next_anon_trees_ids]
                 self.write_candidates_info_to_file(next_anon_trees,
-                                                   f'{IPathFinder.get_file_prefix_by_user_tree(user_anon_tree, self.candidates_file_id)}_same_tree_candidates')
+                                                   f'{self.get_file_prefix_by_user_tree(user_anon_tree, self.candidates_file_id)}_same_tree_candidates')
```
elena-lyulina, line 76: You can create a field with `self.get_file_prefix_by_user_tree(user_anon_tree, self.candidates_file_id)` and use it inside methods instead of explicitly passing it because its value doesn't change like you do with `candidates_file_id` (which can be removed in this way)

src/main/solution_space/path_finder_test_system.py
```diff
@@ -27,13 +27,15 @@
...
 from src.main.solution_space.measured_tree.measured_tree import IMeasuredTree
+from src.main.util.consts import LOGGER_NAME, INT_EXPERIENCE, TEST_RESULT, TASK
 from src.main.solution_space.solution_space_serializer import SolutionSpaceSerializer
 from src.main.solution_space.solution_space_visualizer import SolutionSpaceVisualizer
 
 log = logging.getLogger(LOGGER_NAME)
 
 
 class TEST_INPUT(Enum):
+    INDEX = 'index'
```
elena-lyulina, line 38: Maybe we don't need INDEX here? You change it to  `i` anyway at line 137:
`test_input[TEST_INPUT.INDEX] = i`
so it doesn't store any information.
So according to realization there is no need to pass INDEX together with other test inputs, that's also a reason not to include INDEX in TEST_INPUT.

src/main/solution_space/path_finder_test_system.py
```diff
@@ -130,11 +132,12 @@ def get_result_table(self, title: str) -> Optional[PrettyTable]:
         table = PrettyTable(field_names=[e.value for e in TEST_INPUT] +
```
elena-lyulina, line 132: Instead of adding new value to TEST_INPUT, you can just do 
field_names = ["index"] + [e.value for e in TEST_INPUT]

src/main/solution_space/path_finder_test_system.py
```diff
@@ -130,11 +132,12 @@ def get_result_table(self, title: str) -> Optional[PrettyTable]:
...
                                         [self.__get_path_finder_version(pf) for pf in path_finders], title=title)
 
-        for test_input in self._test_inputs:
+        for i, test_input in enumerate(self._test_inputs):
             user_anon_tree, user_canon_tree = self.__create_user_trees(test_input)
+            test_input[TEST_INPUT.INDEX] = i
```
elena-lyulina, line 137: Just add `i` in the row instead of putting INDEX in the `test_input` dict, the latter is more complicated and hard to understand

src/main/solution_space/path_finder_test_system.py
```diff
@@ -130,11 +132,12 @@ def get_result_table(self, title: str) -> Optional[PrettyTable]:
...
             row = [test_input[key] for key in TEST_INPUT]
             for path_finder in path_finders:
-                row.append(self.__run_path_finder(path_finder, user_anon_tree, user_canon_tree))
+                row.append(self.__run_path_finder(path_finder, user_anon_tree, user_canon_tree, f'{i}'))
```
elena-lyulina, line 140: You can just pass i without casting it to str, python will do it for you

PR #56, opened 2020-05-18
Delete loops from code chains

nbirillo:


src/main/solution_space/solution_space_handler.py
```diff
@@ -155,10 +155,40 @@ def __create_code_info_chain(file: str, task: TASK) -> List[Tuple[Code, CodeInfo
...
+# Before changes: Tree_1 -> Tree_2 -> Tree_3 -> Tree_1 -> Tree_4
+# After changes: Tree_1 -> Tree_4
+def __remove_loops(code_info_chain: List[Tuple[Code, CodeInfo]], user: User) -> List[Tuple[Code, CodeInfo]]:
+    # Add empty code
+    # We want to remove loops, but we can have the situation: Tree1 -> Tree2 -> Tree3 -> Empty Tree -> Tree5
+    # In the case we will get after loops removing: Tree1 -> Tree2 -> Tree3 -> Empty Tree -> Tree5,
+    # but it is incorrect, because we expect Empty Tree -> Tree5
+    # If we add empty code, we will have: Empty Tree -> Tree1 -> Tree2 -> Tree3 -> Empty Tree -> Tree5
+    # and we will get after loops removing: Empty Tree -> Tree5
+    code_info_chain = [(Code.from_source('', TEST_RESULT.CORRECT_CODE.value), CodeInfo(user))] + code_info_chain
```
elena-lyulina, line 176: Please, add todo with a case when there is no empty_code in the chain
Оr maybe add a flag that stores True if loop with empty_code was found (if same_tree_indexes were not empty for empty anon_code). And If it's False, then we should remove the empty_code at the beginning of the chain.

src/test/solution_space/solution_space_handler/remove_loops_test.py
```diff
@@ -0,0 +1,117 @@
...
+from typing import List, Tuple, Callable
+
+import pytest
+
+from src.main.solution_space.data_classes import CodeInfo, User
+from src.main.solution_space.serialized_code import Code
+from src.main.solution_space.solution_space_handler import __remove_loops
+from src.main.util.consts import LOGGER_NAME, TEST_RESULT
+from src.main.util.helper_classes.id_counter import IdCounter
+from src.test.test_config import to_skip, TEST_LEVEL
```
elena-lyulina, line 13: imports order

PR #58, opened 2020-05-20
Bugfix/plots

nbirillo:
- Fix bugs for plots
- Add CLI
- I am not sure about ALGO_LEVEL values. Maybe HINT has to be the default value with -1


elena-lyulina:
Beutiful readme and great work, thanks!

PR #59, opened 2020-05-26
Evaluation handler

nbirillo:
Add a parser for labeled data

### DIFF
--- a/src/main/evaluation/util.py
+++ b/src/main/evaluation/util.py
@@ -3,6 +3,7 @@
 import logging
 from enum import Enum
 from typing import Optional
+from itertools import repeat
 
 from src.main.util import consts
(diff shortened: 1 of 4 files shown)


src/main/evaluation/util.py
```diff
@@ -0,0 +1,155 @@
...
+
+from src.main.util import consts
+from src.main.util.log_util import log_and_raise_error
+
+log = logging.getLogger(consts.LOGGER_NAME)
+
+
+# Check if al least N (min_number_of_true) values are True
+def contains_at_least_n_true(*values: bool, min_number_of_true: int = 2) -> bool:
+    return all(map(any, repeat(iter(list(values)), min_number_of_true)))
```
elena-lyulina, line 16: Let's use sum :)

PR #60, opened 2020-06-04
New cli

nbirillo:
I am not sure about two things:
- What name is better for cli files? Now it contains **main** suffix, but maybe is it better to use **cli** prefix like the **algo_cli**?
- Should I make abstract methods private in the _ICli_? Python does not have a good way to do a private abstract method and I am not sure if should I use it or not


src/main/cli/test_system_main.py
```diff
@@ -0,0 +1,31 @@
...
+        self.__construct_graph()
+
+        # It's possible not to include TEST_INPUT.RATE in dict, in this case it will be found by
+        # running tests on TEST_INPUT.SOURCE_CODE.
+        # However, to speed up the process, one may include TEST_INPUT.RATE.
+        # Todo: get ages and experiences from args?
+        ages = [12, 15, 18]
+        experiences = [INT_EXPERIENCE.LESS_THAN_HALF_YEAR, INT_EXPERIENCE.FROM_ONE_TO_TWO_YEARS,
+                       INT_EXPERIENCE.MORE_THAN_SIX]
+        test_fragments = TestSystem.generate_all_test_fragments(ages, experiences,
```
elena-lyulina, line 24: generate_all_test_fragments -> generate_all_test_inputs

src/main/util/configs.py
```diff
@@ -1,24 +1,34 @@
 # Copyright (c) 2020 Anastasiia Birillo, Elena Lyulina
+
```
elena-lyulina, line 2: Maybe this file also should belong to src/main/cli ?

src/main/util/consts.py
```diff
@@ -271,3 +271,6 @@ class SPLIT(Enum):
 MAX_DIFF_SYMBOLS = 30
 
 TIMEOUT = 5
+
+TRUE_VALUES_SET = {'true', 't', '1', 'yes', 'y'}
+FALSE_VALUES_SET = {'false', 'f', '0', 'no', 'n'}
```
elena-lyulina, line 276: This is also more about cli, maybe we should think about placing it somewhere it src/main/cli? But I'm not sure

src/main/cli/configs.py
```diff
@@ -39,6 +46,20 @@ class PREPROCESSING_LEVEL(Enum):
     INEFFICIENT_STATEMENTS = 4
     INT_EXPERIENCE = 5
 
+    @property
+    def __level_handlers(self):
```
elena-lyulina, line 50: I think `level_actions` is a more proper name because 'handlers' is too general. What do you think?

src/main/cli/configs.py
```diff
@@ -3,6 +3,13 @@
 from enum import Enum
 from typing import List
 
+from src.main.preprocessing.inefficient_statements_removing import remove_inefficient_statements
+from src.main.preprocessing.int_experience_adding import add_int_experience
+from src.main.preprocessing.intermediate_diffs_removing import remove_intermediate_diffs
+from src.main.preprocessing.preprocessing import preprocess_data
+from src.main.splitting.splitting import split_tasks_into_separate_files
+from src.main.splitting.tasks_tests_handler import run_tests
```
elena-lyulina, line 11: Imports order (we definitely need an automatic tool for that or reconsidering our imports style)

src/main/cli/configs.py
```diff
@@ -0,0 +1,135 @@
...
+        return [p_t.value for p_t in PLOT_TYPE]
+
+    @classmethod
+    def description(cls) -> str:
+        return f'{PLOT_TYPE.PARTICIPANTS_DISTRIBUTION.value} - visualize participants distribution; ' \
+               f'{PLOT_TYPE.TASKS_DISTRIBUTION.value} - visualize tasks distribution; ' \
+               f'{PLOT_TYPE.SPLITTING_PLOTS.value} - visualize splitting plots;'
+
+    @classmethod
+    def str_to_plot_type(cls, value: str) -> 'PLOT_TYPE':
```
elena-lyulina, line 34: It's better to use `from __future__ import annotations` instead of writing quotes: it works faster, it's a more modern way to solve a problem with forwarding references since Python 3.7 and it will be used in a future Python releases.

src/main/cli/configs.py
```diff
@@ -0,0 +1,135 @@
...
+    def min_value(cls) -> int:
+        return min([a_t.value for a_t in PREPROCESSING_LEVEL])
+
+    @classmethod
+    def max_value(cls) -> int:
+        return max([a_t.value for a_t in PREPROCESSING_LEVEL])
+
+    @classmethod
+    def description(cls) -> str:
+        return f'the the Nth level runs all the level before it; ' \
```
elena-lyulina, line 73: the the -> the

src/main/cli/plots.py
```diff
@@ -31,63 +40,68 @@ def str_to_participants_distribution_statistics_key(cls, value: str) -> STATISTI
...
+                                             f'Available values: {available_values_message}')
 
     @classmethod
     def str_to_extension(cls, value: str) -> EXTENSION:
         available_extensions = [EXTENSION.HTML, EXTENSION.PNG]
         available_extensions_message = ', '.join(list(map(lambda e: e.value.strip('.'), available_extensions)))
         error_message = f'{value} is not a valid extension. Available values: {available_extensions_message}'
         try:
-            extension = EXTENSION(value.strip('.'))
+            extension = EXTENSION(add_symbol_to_begin(value.rstrip('.'), '.'))
```
elena-lyulina, line 61: What's the reason to rstrip a dot and then add it again? Especially when you check for dot presence in  add_symbol_to_begin()

src/main/cli/preprocessing.py
```diff
@@ -0,0 +1,61 @@
...
+        args = self._parser.parse_args()
+        self._path = self.handle_path(args.path[0])
+        self._level = self.str_to_preprocessing_level(args.level)
+
+    def main(self) -> None:
+        self.parse_args()
+        paths = [self._path]
+        for level_index in range(0, self._level.value + 1):
+            current_level = PREPROCESSING_LEVEL(level_index)
+            self._log.info(f'Current operation is {current_level.level_handler()}')
```
elena-lyulina, line 40: If you agree to change handler to actions, it's better to say 'Current action is ...'

src/main/cli/util.py
```diff
@@ -42,15 +43,36 @@ def str_to_bool(cls, value: any) -> bool:
...
+
+    @classmethod
+    def str_to_preprocessing_level(cls, level: str) -> PREPROCESSING_LEVEL:
+        message = f'Preprosessing level has to be an integer number from {PREPROCESSING_LEVEL.min_value()} ' \
+                  f'to {PREPROCESSING_LEVEL.max_value()}'
+        try:
+            level = int(level)
+            return PREPROCESSING_LEVEL(level)
+        except ValueError:
+            raise argparse.ArgumentTypeError(message)
```
elena-lyulina, line 70: These functions are quite similar, is it possible to pass ALGO_LEVEL or PREPROCESSING_LEVEL as arguments? Also, maybe ICli is the wrong place for it? Not all iCli subclasses have levels, so for me, it is more proper to place it in PREPROCESSING_LEVEL or ALGO_LEVEL. And since these classes are quite similar too, it's better to make an interface for them with other common functions like min_value, max_value, description, and the updated str_to_level function. What do you think?

src/main/cli/util.py
```diff
@@ -42,15 +43,36 @@ def str_to_bool(cls, value: any) -> bool:
...
+                  f'to {ALGO_LEVEL.max_value()}'
+        try:
+            level = int(level)
+            return ALGO_LEVEL(level)
+        except ValueError:
+            raise argparse.ArgumentTypeError(message)
+
+    @classmethod
+    def str_to_preprocessing_level(cls, level: str) -> PREPROCESSING_LEVEL:
+        message = f'Preprosessing level has to be an integer number from {PREPROCESSING_LEVEL.min_value()} ' \
```
elena-lyulina, line 64: Proprosessing -> Preprocessing

src/test/cli/algo/algo_cli_test.py
```diff
@@ -0,0 +1,93 @@
...
+                        [],
+                        [ALGO_PARAMS.DESERIALIZE],
+                        [ALGO_PARAMS.DESERIALIZE, ALGO_PARAMS.SERIALIZE],
+                        [ALGO_PARAMS.DESERIALIZE, ALGO_PARAMS.SERIALIZE, ALGO_PARAMS.VISUALIZE]
+                    ])
+    def param_deserialize_graph(request) -> List[ALGO_PARAMS]:
+        return request.param
+
+    def test_deserialize_graph(self, param_deserialize_graph) -> None:
+        # Construct the graph and after to run the test with it
```
elena-lyulina, line 74: Construct the graph and after to run the test with it -> Construct the graph and then run the test with it

src/main/cli/algo.py
```diff
@@ -1,15 +1,22 @@
 # Copyright (c) 2020 Anastasiia Birillo, Elena Lyulina
 
+import sys
+import logging
+
+sys.path.append('.')
```
elena-lyulina, line 6: Did you try to run it right from the src/main/cli path by calling `python algo.py <args>`?

nbirillo: Yes, it works on my laptop (I have macOS). I tried to run it from:
- the `root` project
- the `src` folder
- the `main` folder
- the `cli` folder

PR #64, opened 2020-08-19
Preprocess new data

nbirillo:
- Merge ide-events files
- Delete unnecessary code tracker files
- Delete files with test mode = ON


src/main/preprocessing/code_tracker_handler.py
```diff
@@ -10,24 +10,20 @@
...
-    values = np.delete(values, index)
-    # If list is empty after removing defaults, return default value
-    if values.size == 0:
-        return default_value.value
-    # If we have only 1 valid element after removing defaults, we should return it
-    elif len(values) == 1 and fits_column_restriction(values[0]):
+    # Todo: make it more readable and delete not necessary params
+    # Delete all possible default values. If we have only 1 valid element after removing defaults, we should return it
+    # Otherwise it is INVALID_FILE_FOR_PREPROCESSING
+    values = [x for x in values if x not in [np.nan, None, np.datetime64('NaT'), 0] and not pd.isna(x)]
```
elena-lyulina, line 24: I can be wrong, but it seems that new data can have -1 in the 'age' column, for example, because it's an initial value. So it's okay to have -1 and the actual value in the 'age' column, those files shouldn't be invalid
image

src/main/preprocessing/preprocessing.py
```diff
@@ -1,103 +1,72 @@
...
-            if count_ati >= 2:
-                log_and_raise_error('The number of activity tracker files is more than 1', log)
-    return ati_index
-
-
-def __has_files_with_same_names(files: List[str]) -> bool:
-    original_files = list(map(get_original_file_name, files))
-    return len(original_files) != len(set(original_files))
+def __partition_into_ct_and_ati_files(files: List[str]) -> Tuple[List[str], List[str]]:
+    ati_files = [f for f in files if 'ide-events' in f]
```
elena-lyulina, line 19: let's use a variable ACTIVITY_TRACKER_FILE_NAME instead of 'ide-events' (we already have it in consts)

src/main/preprocessing/preprocessing.py
```diff
@@ -1,103 +1,72 @@
...
+            return True
+    return False
 
-            write_result(output_directory, path, ct_file, ct_df)
 
-        log.info(f'Finish handling the folder {folder}')
+# Merge ide-events files
+# Delete unnecessary code tracker files
+# Delete files with test mode = ON
+def preprocessing(path: str) -> str:
```
elena-lyulina, line 54: let's use **verbs** in function names

src/main/preprocessing/preprocessing.py
```diff
@@ -1,103 +1,72 @@
...
+    output_directory = get_output_directory(path, consts.PREPROCESSING_DIRECTORY)
+    user_folders = get_all_file_system_items(path, lambda dir: 'user' in dir, consts.FILE_SYSTEM_ITEM.SUBDIR)
+    for user_folder in user_folders:
+        user_path = os.path.join(path, user_folder)
```
elena-lyulina, line 58: user_folder is already a path, why you want to join it? Btw, it works since user_folders starts with the root directory, so ps.path.join doesn't join it with path

src/main/preprocessing/preprocessing.py
```diff
@@ -1,103 +1,72 @@
...
+        output_user_path = os.path.join(output_directory, get_name_from_path(user_folder, False))
+        log.info(f'Start handling the path {user_path}')
+        task_folders = get_all_file_system_items(user_path, all_items_condition, consts.FILE_SYSTEM_ITEM.SUBDIR)
+        for task_folder in task_folders:
+            task_path = os.path.join(user_path, task_folder)
```
elena-lyulina, line 63: again, you don't need to join it

src/main/preprocessing/preprocessing.py
```diff
@@ -1,103 +1,72 @@
...
+            output_task_path = os.path.join(output_user_path, get_name_from_path(task_folder, False))
+            log.info(f'Start handling the folder {task_path}')
+            files = get_all_file_system_items(task_path, extension_file_condition(EXTENSION.CSV))
+            ct_files, ati_files = __partition_into_ct_and_ati_files(files)
+            if __handle_ct_files(ct_files, output_task_path):
+                if ati_files:
```
elena-lyulina, line 69: let's combine these two conditions

elena-lyulina (changes requested):
You renamed src/main/cli/preprocessing.py → src/main/cli/merging_ct_with_ati.py, but you didn't rename PreprocessingCli and other variables in this file. Please, finish the renaming.
Everything else looks great, good job :)

nbirillo:
Sorry, It was my mistake - in CLI module I had not to rename _src/main/cli/preprocessing.py_ → _src/main/cli/merging_ct_with_ati.py_, because this module is a general module for all preprocessing steps

PR #66, opened 2020-08-20
New structure - splitting data

nbirillo:


src/main/splitting/splitting.py
```diff
@@ -71,21 +71,36 @@ def find_task_dfs(df: pd.DataFrame, task: consts.TASK) -> List[pd.DataFrame]:
...
+# --user_N2
+# ---task1
+# ----user_N2_files
+# After the function calling:
+# -root
+# --task1
+# ----user_N1_files
+# ----user_N2_files
+# Todo: I am not sure about this name
 def split_tasks_into_separate_files(path: str, output_directory_suffix: str = 'separated_tasks') -> str:
```
elena-lyulina, line 97: Let's call it 'reorganize_files_structure'? And maybe keep the previous function just in case as a separate one?

elena-lyulina (changes requested):
Almost done!

src/main/splitting/splitting.py
nbirillo: I think we can find the previous version in the commits history if we want

PR #68, opened 2020-08-24
Remove intermediate diffs for the new data structure

nbirillo:


src/main/util/file_util.py
```diff
@@ -303,3 +303,7 @@ def copy_file(src: str, dst: str) -> None:
     if is_file(src):
         create_directory(get_parent_folder(dst))
         copyfile(src, dst)
+
+
+def move_directory(src: str, dst: str) -> None:
+    shutil.move(src, dst)
```
elena-lyulina, line 309: Why do you need this method?

elena-lyulina (changes requested):

src/main/util/file_util.py
nbirillo: You are right, I forgot to delete this function. Thanks!

elena-lyulina (approved):
Great!

PR #69, opened 2020-08-25
Add anonymize_code function for python

nbirillo:
- Update inefficient_statements_removing
- Add anonymize_code for Python


elena-lyulina (approved):
Nice :+1:

PR #71, opened 2020-08-25
Fix get_profile_statistics function for the new data structure

nbirillo:


src/main/statistics_gathering/statistics_gathering.py
```diff
@@ -44,10 +46,32 @@ def __get_ct_df(ct_file: str, needs_handling: bool = True) -> pd.DataFrame:
     return pd.read_csv(ct_file, encoding=consts.ISO_ENCODING)
 
 
+def __get_experience(ct_df: pd.DataFrame) -> Profile:
+    if STATISTICS_KEY.EXPERIENCE.value not in ct_df.columns:
+        # New data structure
+        experience_years = __get_profile_info(ct_df, STATISTICS_KEY.EXPERIENCE_YEARS)
+        experience_months = __get_profile_info(ct_df, STATISTICS_KEY.EXPERIENCE_MONTHS)
+        if 1 <= experience_months < 6:
```
elena-lyulina, line 54: shouldn't it be `0 <= experience_months < 6` ?

elena-lyulina (changes requested):

src/main/statistics_gathering/statistics_gathering.py
nbirillo: I think no because `experience_months` can not be zero. Or not?

elena-lyulina: It can, I think, be any number in [0, 11]

elena-lyulina (approved):
Cool

PR #72, opened 2020-08-27
Fixing plots timestamps

elena-lyulina:


src/main/util/file_util.py
```diff
@@ -68,8 +68,11 @@ def get_name_from_path(path: str, with_extension: bool = True) -> str:
 
 # Not empty extensions are returned with a dot, for example, '.txt'
 # If file has no extensions, an empty one ('') is returned
-def get_extension_from_file(file: str) -> EXTENSION:
-    return EXTENSION(os.path.splitext(file)[1])
+def get_extension_from_file(file: str) -> Optional[EXTENSION]:
```
nbirillo, line 71: Please, change the function **get_name_from_path** accoding to new **get_extension_from_file** function. I mean, if the **get_extension_from_file** function returns **None**,  in the function **get_name_from_path**  you should not return something like **name.None**

nbirillo (changes requested):
Great job, but we should fix tests before merging

PR #73, opened 2020-11-10
Prepare for paper: preprocessing

nbirillo:
Add some comments for the preprocessing step, check if it works, fix readme and fix CLI for preprocessing.


src/main/preprocessing/preprocessing.py
```diff
@@ -23,6 +23,9 @@ def __partition_into_ct_and_ati_files(files: List[str]) -> Tuple[List[str], List
 
 
 def __merge_ati_files(ati_files: List[str]) -> pd.DataFrame:
+    """
+    Combine all active tracker files according to timestamps, excluding duplicates.
```
elena-lyulina, line 27: active -> activity

src/main/preprocessing/preprocessing.py
```diff
@@ -34,11 +37,23 @@ def __merge_ati_files(ati_files: List[str]) -> pd.DataFrame:
...
-# The function returns True if the code tracker file was created and False otherwise
 def __handle_ct_files(ct_files: List[str], output_task_path: str) -> bool:
+    """
+    The function returns True if new code tracker file was created and False otherwise
+    We should choose the last state of the code tracker files for the task or all last states and create a new file
+    where we union them. The student can submit the solution several times, while the history of the codetracker file
+    is not erased. In this way, we only need to select the final file with the entire history. On the other hand,
+    if the file was full, then it will be sent additionally and new files will contain a new history.
+    In this case, it is necessary to find the last states of all files with a unique history, combine according to
+    timestamps and write to a new final file.
```
elena-lyulina, line 54: hmmm can we combine files that are full?

src/main/splitting/tasks_tests_handler.py
```diff
@@ -101,6 +101,19 @@ def __get_user_folder_name_from_path(file: str) -> str:
...
+    Note: the enum class TASK (see consts.py file)  must have the task key.
+    It also must match the name of the folder with test files in the resources/tasks_tests.
+
+    For example, if your task has key [my_key], you should add a new value into TASK const with value [my_key]
+    and add a new folder [my_key] with input and output files for tests in the resources/tasks_tests folder.
+
+    The test result is an array containing values for all tasks from the TASK enum class.
+    If the code snapshot is incorrect, then the value -1 is specified.
+    To deserialize this array of ratings, use the function unpack_tests_results from splitting.py.
+    TODO: should we add a function to get the tests results by the current task?
```
elena-lyulina, line 115: I thought we already have had such a function, but now i'm not sure
Maybe we shouldn't run tests for all tasks now since we improved codetracker and it's quite impossible to have a solution with the wrong task

elena-lyulina (changes requested):
Thanks a lot for adding comments, I barely remembered what was the purpose of all these functions.

src/main/preprocessing/merging_ct_with_ati.py
```diff
@@ -12,13 +12,18 @@
...
-    get_parent_folder_name, get_output_directory, write_result, extension_file_condition, user_subdirs_condition
+from src.main.util.file_util import get_original_file_name, get_all_file_system_items, get_output_directory, \
+    write_result, extension_file_condition, user_subdirs_condition
 
 log = logging.getLogger(consts.LOGGER_NAME)
 
 
 def is_ct_file(csv_file: str, column: consts.CODE_TRACKER_COLUMN = consts.CODE_TRACKER_COLUMN.CHOSEN_TASK) -> bool:
+    """
+    Check if the file with a name that has the activity tracker file name is codetracker file.
```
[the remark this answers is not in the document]
nbirillo: sorry, it is my mistake

src/main/preprocessing/preprocessing.py
nbirillo: I added todo: `# TODO: handle the case with additional files`

I think we should do it in the next PR. But if you want I can do it now

src/main/splitting/tasks_tests_handler.py
nbirillo: We run tests only for the current task, but store array for all tasks with 0 value for other tasks.

nbirillo: Also, we have a function to unpack tests results into an array, but not for getting score for the current task

src/main/preprocessing/preprocessing.py
elena-lyulina: Let's leave it till the next PR

elena-lyulina (approved):
Great!

